PR #6, opened 2025-12-02
Save error messages

Sachanski:
* When an error occurs while saving if the error is a more concrete APIError (i.e. not a ServerError`) the corresponding error message is displayed
* The backend's error handler now transforms `PayloadTooLarge` errors (from express) into a `BadRequestError`s


client/src/pages/tapestry/view-model/tapestry-data-sync.ts
```diff
@@ -111,7 +112,14 @@ export class TapestryDataSync {
           (model) => {
             --model.pendingRequests
           },
-          !!error && setSnackbar({ variant: 'error', text: 'Error during save' }),
+          !!error &&
+            setSnackbar({
+              variant: 'error',
+              text:
+                error instanceof APIError && error.data.name !== 'ServerError'
+                  ? error.message
```
rganchev: I'm not sure it is a good idea to display server error messages here. They are not meant to be user-friendly. Additionally, in this particular case, the message doesn't tell the user what is the actual restriction.

rganchev (approved):

PR #10, opened 2025-12-04
Path style option

Sachanski:
* Added an env var that allows for changing the virtual host style to path style for S3 urls


server/src/config.ts
```diff
@@ -3,6 +3,10 @@ import z from 'zod/v4'
 import { deepFreeze } from 'tapestry-core/src/utils.js'
 
 const port = z.coerce.number<number>().int().min(0).max(65535)
+const assertTrue = z
```
rganchev: The word `assert` hints that something would break if this value isn't `true`. Isn't this just "boolean"?

rganchev (dismissed):

Issue #21, opened 2025-12-10
Crash on mobile Safari


Sachanski:
On tapestries with many elements Safari draw excessive compositing layers which leads to crashes (OOM) for example when zooming from a viewport showing all items to a particular item or group

Issue #20, opened 2025-12-10
Info dialog not working


rganchev:
Trying to open the item info dialog in any tapestry on dev displays an error.

PR #22, opened 2025-12-10
[#21, #20] Safari optimizations

Sachanski:
* Added a `will-change: transform;` css property to the dom-container. The idea is to force the browser to keep the compositing layer even when the element is outside the viewport.
* The collaborator containers are not rendered when in viewer mode.
* Hiding the tapestry elements when they are not visible in the viewport. This leads to a better performance when we have a lot of elements and are zooming in-out while close to the maximum scale
* Replaced lodash with lodash-es on the client and fixed a lodash import in `core-client`
* Renamed the `Toolbars` component to `MainToolbar` and removed an extra div
* The presentation order canvas is with `display: none;` instead of visiblity hidden (again in order to prevent the drawing of the extra compositing layer). Added a `this.stage.pixi.presentationOrder.queueResize` invocation in order to size the canvas properly when is is displayed
* Removed a usage of lodash's `chain` funciton since due to the tree shaking it did not include the reset of the lodash functions used in the chain

### DIFF
--- a/client/src/pages/tapestry/view-model/store-commands/tapestry.ts
+++ b/client/src/pages/tapestry/view-model/store-commands/tapestry.ts
@@ -1,10 +1,5 @@
-// <<<<<<< HEAD
-// import { omit, sample } from 'lodash-es'
-// import { ORIGIN, Point } from 'tapestry-core/src/lib/geometry'
-// =======
 import { omit, sample } from 'lodash-es'
 import { maxEmptyArea, ORIGIN, Point, Rectangle } from 'tapestry-core/src/lib/geometry'
-// >>>>>>> main
 import { TapestryDto } from 'tapestry-shared/src/data-transfer/resources/dtos/tapestry'
 import { EditableTapestryViewModel, InteractionMode, IAImport, convertCommand } from '..'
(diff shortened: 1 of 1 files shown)


core/src/lib/geometry.ts
```diff
@@ -133,6 +133,15 @@ export class Rectangle {
     })
   }
 
+  isEqual(other: Rectangle) {
+    return (
+      this.position.x === other.position.x &&
+      this.position.y === other.position.y &&
+      this.size.width === other.size.width &&
+      this.size.height === other.size.height
```
rganchev: I wonder if it makes sense to compare `left`, `top`, `width`, and `height` instead. This way `new Rectangle({ x: 0, y: 0 }, { width: 1, height: 1 })` would be equal to `new Rectangle({ x: 1, y: 1 }, { width: -1, height: -1 })`, for example, which sounds logical - they are exactly the same shape. I know we may not have negative-size rectangles at the moment, but the problem is conceptual, since nothing is explicitly forbidding them.

rganchev (dismissed):

rganchev (approved):

PR #28, opened 2025-12-16
Multiple file import

Sachanski:
* When importing files via the import toolbar button multiple files can now be chosen
  * This led to a small refactoring in order to consolitdate all places (`PasteButton`, `ImportToolbar`, `EditorGlobalEventsController`) where files are being transfered (pasted/dropped/imported)
* Added two new item factories for parsing .webloc and .url files (these are the page shortcuts created when dragging a page to the file system from MacOS and Windows respectively). Because of these there are now two new dependencies - one for determining the file type and one for parsing and .ini file (the Windows page shortcuts)
* Fixed an issue where the custom color picker menu was closed immediatelly after changing the value when working with text items and action buttons
* The redis service now allows its clients to overwrite the cache ttl based on the result of the async action. This is useful for the wayback machine snapshot requests where we would like to cache pages that are not yet indexed for a shorter period.

### DIFF
--- a/client/src/stage/item-factories.ts
+++ b/client/src/stage/item-factories.ts
@@ -66,7 +66,7 @@ async function parseWeblocFile(source: File) {
  * An ItemFactory takes a MediaItemSource (File or URL) and tries to produce one or more tapestry items from it.
  * If a factory doesn't know how to handle a given source, it returns null.
  */
-export type ItemFactoryResult = [ItemCreateDto[], IAImport[]]
+export type ItemFactoryResult = { items: ItemCreateDto[]; iaImports: IAImport[] }
 type ItemFactory = (
   source: MediaItemSource,
   mediaType: string | null,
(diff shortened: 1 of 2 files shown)


client/src/stage/item-factories.ts
```diff
@@ -89,6 +91,52 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
...
+
+  type URLSection = Record<'URL', string> | undefined
+  const url = (parse(await source.text()).InternetShortcut as URLSection)?.URL
+  if (!url) {
+    return null
+  }
+
+  return (
+    (await iaCollectionFactory(url, _, tapestryId)) ??
+    (await webpageItemFactory(url, _, tapestryId))
```
rganchev: Why are you only considering IA collections and webpages here? What would happen if the file contains an image URL, for example? Shouldn't we iterate over all item factories like we do in `parseFileTransferData`?

rganchev (dismissed):

PR #52, opened 2026-03-17
Disable Chrome cache to save disk space

rganchev:
Chrome creates "user data" directories on every run and stores various caches inside them. We have little use of them because of the way we use Puppeteer. They build up in time and consume a lot of disk space. I have disabled them here in order to save disk space.

### DIFF
--- a/server/src/tasks/thumbnail-generators/webpage.ts
+++ b/server/src/tasks/thumbnail-generators/webpage.ts
@@ -1,4 +1,7 @@
 import axios from 'axios'
+import { mkdtemp, rm } from 'node:fs/promises'
+import { join } from 'node:path'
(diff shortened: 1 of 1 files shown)


Sachanski (approved):

rganchev:
While debugging memory problems on our dev environment, I found a large directory of Chrome user data and thought that the memory problems would be mitigated by the fix in this PR. However, later, the memory problems were associated with Docker cache issues and were fixed in a different manner, so at this point I am not certain whether this PR actually fixes anything. Until I verify whether it does, I prefer not to merge it. If I can't find an evidence that Chrome actually leaks user data and consumes disk space, I would rather remove this fix than risk merging it and breaking something else.

PR #59, opened 2026-05-18
Disable optimisations in edit mode

zmarinov-astea:


core-client/src/view-model/index.ts
```diff
@@ -79,6 +79,7 @@ export interface ItemViewModel<I extends Item = Item> {
   readonly snapshotId?: string | null
   readonly hasBeenActive?: boolean
   readonly isPlaying?: boolean
+  readonly hasRendered?: boolean
```
zmarinov-astea: Maybe has rendered is not a very appropriate name, since an element gets rendered if it doesn't have a snapshot yet, but it gets hidden after the snapshot gets created. In other words we don't need to preserve its internal state.

Sachanski (approved):

PR #60, opened 2026-05-22
Move comments indicator to pixi

zmarinov-astea:
1. Created CommentsIndicatorContainer class for pixi rendering.
2. Deleted the old comments indicator component.

### DIFF
--- a/core-client/src/components/tapestry/items/tapestry-item/styles.module.css
+++ b/core-client/src/components/tapestry/items/tapestry-item/styles.module.css
@@ -11,14 +11,10 @@
   .title-container {
     position: absolute;
     bottom: calc(100% + var(--title-margin));
-    width: 100%;
+    width: 80%;
     padding: 0 4px;
     box-sizing: border-box;
     color: var(--theme-text-primary);
-    display: flex;
-    gap: 8px;
-    justify-content: space-between;
-    align-items: end;
   }
 
   .wrapper {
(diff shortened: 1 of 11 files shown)


core-client/src/components/tapestry/items/tapestry-item/styles.module.css
```diff
@@ -11,14 +11,10 @@
...
     bottom: calc(100% + var(--title-margin));
-    width: 100%;
+    width: 80%;
     padding: 0 4px;
     box-sizing: border-box;
     color: var(--theme-text-primary);
-    display: flex;
-    gap: 8px;
-    justify-content: space-between;
-    align-items: end;
```
Sachanski, line 21: This was some leftover from a previous iteration I assume?

core-client/src/components/tapestry/items/tapestry-item/styles.module.css
```diff
@@ -11,14 +11,10 @@
   .title-container {
     position: absolute;
     bottom: calc(100% + var(--title-margin));
-    width: 100%;
+    width: 80%;
```
Sachanski, line 14: The 80% are to leave some space for the comment indicator I assume?

Sachanski (approved):

core-client/src/components/tapestry/items/tapestry-item/styles.module.css
zmarinov-astea: The title and the comment indiactor were in one flex container

core-client/src/components/tapestry/items/tapestry-item/styles.module.css
zmarinov-astea: Yes, although we would need to discuss again the purpose of the title and maybe draw it on pixi as well, since now it is not visible when the element is not interactive...

Issue #30, opened 2025-12-18
Zen mode


rganchev:
Implement some functionality that allows viewers to hide all on-screen menus so they can focus better on the content of the Tapestry. This may also be useful when presenting Tapestries on kiosks, navigating through a presentation, etc.

TBD:
 - How should this mode be enabled/disabled?
 - Should item halos also be hidden or just the static toolbars?

PR #61, opened 2026-05-27
Remove shadows from item snapshots

zmarinov-astea:
Added a new hideControls property, which hides all tapestry page toolbars while preserving zoom toolbar keyboard shortcuts function.
This could be used in #30.


Sachanski (approved):
Why is showing the zoom toolbar specifically handled in its own render function, compared to the rest of the toolbars where the parent component renders them conditionally?

zmarinov-astea:
> Why is showing the zoom toolbar specifically handled in its own render function, compared to the rest of the toolbars where the parent component renders them conditionally?

The sole reason is to keep the keyboard shortcuts for navigation. Radan said that, in the future we will extend this functionality in order to implement the Zen mode. I think it would be better to have access to these shortcuts in Zen mode. Also we could use hideControls=1 parameter for testing/debugging...

PR #63, opened 2026-05-28
Fix pdf unfocus

### LINKED ISSUE #19: Unfocus not working properly
The "unfocus" functionality doesn't work in some cases. Steps to reproduce:

1. Open a tapestry with a PDF, preferably a larger one, so it loads slowly.
2. Select the PDF element in the tapestry to activate it. A loading spinner appears and the PDF starts loading.
3. Before the loading spinner disappears, hit the `F` key or click the "Focus" button in the element's halo. The element is focused.
4. Wait for the PDF to load fully and the loading spinner to disappear.
5. Try to press the `F` key again (or the "Focus" button in the halo).

Actual result: Nothing happens
Expected result: The element is "unfocused", i.e. the viewport state is reset to exactly what it was before step 3 above.

zmarinov-astea:
After the pdf loads a new toolbar menu item gets added for browsing pages, which causes the remount of all menu items. That is why a key should be used to preserve the focus button's state.

### DIFF
--- a/core-client/src/components/lib/toolbar/index.tsx
+++ b/core-client/src/components/lib/toolbar/index.tsx
@@ -20,6 +20,7 @@ export type SubmenuIds<T, D extends number = 10> = [D] extends [never]
 
 export interface ToolbarElement {
   element: ReactElement
+  key?: string
   tooltip?: TooltipProps
   badge?: string | boolean
 }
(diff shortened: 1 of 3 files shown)


core-client/src/components/lib/toolbar/index.tsx
```diff
@@ -94,7 +95,7 @@ function ToolbarRow({ items, selectedSubmenu }: ToolbarRowProps) {
                 />
               </div>
             ) : (
-              <SimpleMenuItem key={index} ui={item} />
+              <SimpleMenuItem key={(isToolbarElement(item) && item.key) || index} ui={item} />
```
Sachanski, line 98: Maybe `??` instead of `||`

Sachanski (approved):

core-client/src/components/lib/toolbar/index.tsx
zmarinov-astea: We always need to fallback to index. If we use ?? and isToolbarElement returns false, then the key will be set to false, which we don't need

PR #65, opened 2026-06-04
Implement webpage to PDF conversion

zmarinov-astea:
1. Created a convert to PDF worker task, which uses puppeteer to convert webpages to PDF
2. Added a button to the webpage toolbar for conversion to PDF. Currently an item update with data { type: 'pdf' } is sent to the server, but we could implement a separate endpoint for conversions.
3. The webpage item gets deleted and replaced by the newly created PDF item.
4. The worker now uses the SocketServer in order to send the tapestry update notification.

### DIFF
--- a/client/src/components/tapestry-elements/items/webpage/index.tsx
+++ b/client/src/components/tapestry-elements/items/webpage/index.tsx
@@ -79,7 +79,7 @@ function Webpage({ src, onLoad, ...props }: WebFrameSwitchProps) {
     <div className={styles.error}>
       <Icon icon="sentiment_very_dissatisfied" />
       <Text>{`Cannot frame ${src}`}</Text>
-      <Text>{`Switch to wayback machine version or convert to PDF.`}</Text>
+      <Text>Switch to wayback machine version or convert to PDF.</Text>
     </div>
   ) : null
 }
--- a/client/src/components/tapestry-elements/items/webpage/index.tsx
+++ b/client/src/components/tapestry-elements/items/webpage/index.tsx
@@ -31,6 +31,7 @@ import {
   WebFrame,
   WebFrameSwitchProps,
 } from 'tapestry-core-client/src/components/tapestry/items/webpage/web-frame'
+import { useAsyncAction } from 'tapestry-core-client/src/components/lib/hooks/use-async-action'
 
 const checkedSources = new Map<string, boolean>()
 
@@ -78,6 +79,7 @@ function Webpage({ src, onLoad, ...props }: WebFrameSwitchProps) {
     <div className={styles.error}>
       <Icon icon="sentiment_very_dissatisfied" />
       <Text>{`Cannot frame ${src}`}</Text>
+      <Text>{`Switch to wayback machine version or convert to PDF.`}</Text>
     </div>
   ) : null
 }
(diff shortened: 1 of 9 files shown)


client/src/components/tapestry-elements/items/webpage/index.tsx
```diff
@@ -78,6 +79,7 @@ function Webpage({ src, onLoad, ...props }: WebFrameSwitchProps) {
     <div className={styles.error}>
       <Icon icon="sentiment_very_dissatisfied" />
       <Text>{`Cannot frame ${src}`}</Text>
+      <Text>{`Switch to wayback machine version or convert to PDF.`}</Text>
```
Sachanski: No need for the interpolation here

server/src/tasks/convert-to-pdf.ts
```diff
@@ -0,0 +1,77 @@
...
+      const s3Key = tapestryKey(item.tapestryId, `${crypto.randomUUID()}.pdf`, true)
+      await tx.item.create({
+        data: {
+          ...pick(item, CONVERT_ITEM_PROPS),
+          type: 'pdf',
+          source: s3Key,
+          scheduledThumbnailProcessing: 'derive',
+        },
+      })
+      await s3Service.putObject(s3Key, value, 'application/pdf')
```
Sachanski, line 66: I am sure there is a reason, but why aren't we updating the item and instead deleting the old and creating a new one. Things like comments/rels will probably get deleted this way?

server/src/tasks/worker.ts
```diff
@@ -12,9 +14,12 @@ async function processTask(job: Job<JobTypeMap[JobName], void, JobName>) {
       return s3Cleanup()
     case 'create-tapestry':
       return createTapestry(job.data as JobTypeMap['create-tapestry'])
+    case 'convert-to-pdf':
+      return convertToPdf(job.data as JobTypeMap['convert-to-pdf'])
   }
 }
 
+void socketServer.init()
```
Sachanski: I am not a big fan of this usage of the socket server. In essence what we want is to notify the database that something happened (the tapestry was updated), so that the application servers can propagate this information through the sockets.

Therefore we only need the `notify` functionality from the `SocketServer` object. It will probably be better to create a local db subscriber and also release it after the notification has been sent, so that we don't keep unnecessary connections opened to the database

server/src/tasks/convert-to-pdf.ts
```diff
@@ -0,0 +1,82 @@
...
+    await scheduleTapestryThumbnailGeneration(item.tapestryId)
+
+    const dbSubscriber = new DBSubscriber()
+    await dbSubscriber.init()
+    await dbSubscriber.notify({
+      name: 'tapestry-updated',
+      tapestryId: item.tapestryId,
+      deletedIds: { items: [item.id] },
+    })
+    await dbSubscriber.close()
```
Sachanski: Might be a bit more concise if there were a `fireNotification` static method on the DBSubscriber class that encapsulates this, so the DBSubscriber client wouldn't need to bother with all the boilerplate around the setup/teardown

Sachanski (dismissed):

Sachanski (approved):

PR #69, opened 2026-06-08
Fix stopping playback after focus

### LINKED ISSUE #68: Video stops playing after focus
When you click on the focus button of a video item or web video that is playing, the video stops and returns to frame 1.

zmarinov-astea:
z.coerce.number used in NullishInt converts null to 0, and searchParams.get('...') returns null when the param is missing, however, in that case the parsed value should be undefined.

### DIFF
--- a/core-client/src/components/tapestry/hooks/use-media-params.ts
+++ b/core-client/src/components/tapestry/hooks/use-media-params.ts
@@ -1,5 +1,5 @@
 import { useSearchParams } from 'react-router'
-import { Id, NullishInt } from 'tapestry-core/src/data-format/schemas/common'
+import { Id, OptionalInt } from 'tapestry-core/src/data-format/schemas/common'
 
 export interface MediaParams {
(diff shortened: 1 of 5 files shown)


core-client/src/components/tapestry/hooks/use-media-params.ts
```diff
@@ -16,8 +16,8 @@ export function useMediaParams(id: Id): MediaParams {
   const autoplay = searchParams.get('autoplay')
 
   return {
-    startTime: NullishInt().safeParse(searchParams.get('startTime')).data,
-    stopTime: NullishInt().safeParse(searchParams.get('stopTime')).data,
+    startTime: NullishInt().safeParse(searchParams.get('startTime') ?? undefined).data,
```
zmarinov-astea: Maybe we should instead change NullishInt to parse null as null not as 0.

Sachanski (approved):

PR #71, opened 2026-06-17
Remove zod schemas from render and smoothen continuous zoom

zmarinov-astea:
The computeRestricetedScale function validated all items through zod schemas and worsened render performance. Even after removing zod schema validation, the function is throttled for optimal results. 
The rate of events fired when pressing +/- keys for continuous zoom was too fast, which caused the animation to constantly stop and start again with new parameters. That is why I created a new single exponential animation which plays only once while the events are being fired until activity ceases for at least 0.1s.

### DIFF
--- a/core-client/src/view-model/store-commands/viewport.ts
+++ b/core-client/src/view-model/store-commands/viewport.ts
@@ -221,7 +221,7 @@ export function zoomOut(continuous = false): StoreMutationCommand<TapestryViewMo
     const scale = viewport.transform.scale
     const minScale = getMinScale(viewport, idMapToArray(items))
     const { zoomStep, animate } = getZoomParameters(scale, minScale, continuous)
-    store.dispatch(transformViewport(zoomToCenter(store.get(), -zoomStep), animate))
+    store.dispatch(transformViewport(zoomToCenter(store.get(), zoomStep), animate))
     if (continuous) {
       continuousZoom = 'ZOOM-OUT'
       stopContinuosZoom()
--- a/core-client/src/view-model/store-commands/viewport.ts
+++ b/core-client/src/view-model/store-commands/viewport.ts
@@ -1,6 +1,6 @@
-import { Tween, Easing } from '@tweenjs/tween.js'
+import { Tween } from '@tweenjs/tween.js'
 import { AnimationOptions, tween } from '../tweening.js'
-import { TapestryViewModel, ZOOM_STEP, MAX_INITIAL_SCALE, MAX_SCALE } from '../index.js'
+import { TapestryViewModel, MAX_INITIAL_SCALE, MAX_SCALE } from '../index.js'
 import {
   LinearTransform,
   Vector,
@@ -23,13 +23,13 @@ import {
   getTranslationRange,
   getMinScale,
   getGroupMembers,
+  getZoomParameters,
 } from '../utils.js'
 import { idMapToArray, pickById } from 'tapestry-core/src/utils.js'
 import {
   cubicBezierPoly,
   Exponent,
   integrate,
-  log,
   Polynomial,
   RungeKutta4,
 } from 'tapestry-core/src/lib/algebra.js'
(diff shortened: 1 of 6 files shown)


core-client/src/view-model/store-commands/viewport.ts
```diff
@@ -187,17 +197,49 @@ export function resizeViewport(size: Size): StoreMutationCommand<TapestryViewMod
...
+    const scale = viewport.transform.scale
+    const minScale = getMinScale(viewport, idMapToArray(items))
+    const zoomStep = continuous ? Math.log(scale / minScale) : ZOOM_STEP
+    const animate = continuous
+      ? {
+          easing: Easing.Linear.None,
+          duration: log(CONTINUOUS_ZOOM_SPEED, scale / minScale),
+          zoomEffect: 'exponential' as const,
+        }
+      : true
```
Sachanski: Feels like we can extract a function for that receives the store and the zoom direction and returns the animate options + the zoom step

core-client/src/view-model/store-commands/viewport.ts
```diff
@@ -27,24 +27,29 @@ import {
...
-const CONTINUOUS_ZOOM_STEP = 0.15
-const CONTINUOUS_ZOOM_ANIMATION_OPTIONS: AnimationOptions = {
-  easing: Easing.Linear.None,
-  duration: 0.1,
-}
+const CONTINUOUS_ZOOM_SPEED = 3
 const ELEMENT_TOOLBAR_PADDING = 65
 
 let zoomAnimation: Tween | undefined = undefined
+let continuousZoom: 'ZOOM-IN' | 'ZOOM-OUT' | null = null
```
Sachanski, line 44: I am not sure that I feel too good about adding more global variables, but can't suggest anything better either.

core/src/utils.ts
```diff
@@ -17,16 +17,26 @@ export function isHTTPURL(str: string | null | undefined): str is `http${string}
...
 }
 
 export function isRel(obj: unknown): obj is Rel {
-  return RelSchema.safeParse(obj).success
+  return isObject(obj) && typeof obj.id === 'string' && isObject(obj.from) && isObject(obj.to)
 }
 
 export function isMediaItem(item: unknown): item is MediaItem {
-  return MediaItemSchema.safeParse(item).success
+  return isObject(item) && typeof item.source === 'string' && isItem(item)
```
Sachanski: I guess right now it is technically true that any item with a source property of type string is a media item, but the real check should be based on the item's type.

Sachanski (commented):
Do we really need to ditch the zod schemas in orderd for this optimization to take effect? I thought that the throttling should be enough.

zmarinov-astea:
> Do we really need to ditch the zod schemas in orderd for this optimization to take effect? I thought that the throttling should be enough.

No, even with throttling, there is a noticeable decline in performance. Moreover, the isItem function is used at other various places throughout the render logic, and it doesn't make sense to pass items through zod schemas on every render

core-client/src/view-model/store-commands/viewport.ts
zmarinov-astea: Another option is to keep this state in the view model

core/src/utils.ts
zmarinov-astea: Ok, do you think we need to check the type in the isItem function as well, for consistency?

Sachanski: If you feel like it I am OK with adding the additional check, altough since we are aiming for performance I'd be fine without it. By the way is the `isMediaItem` function also part of the rendering logic? If not maybe we can keep that schema at least for this one?

zmarinov-astea: It's not going to be consistent to have a schema on one place but not the other. It's not used during rendering, but various other functions use it like when typing in the search box

Sachanski (dismissed):

Sachanski (dismissed):

Sachanski (approved):

PR #72, opened 2026-06-19
Allow PDF conversion only for certain webpage types

zmarinov-astea:
1. Removed PDF conversion button for playable media webpages and added it to wayback page type. 
2. Added some patch validation to the TapestryUndoStack so the application doesn't crash when clicking undo.


client/src/pages/tapestry/view-model/undo-stack/tapestry-undo-stack.ts
```diff
@@ -47,8 +48,47 @@ export class TapestryUndoStack extends BaseUndoStack<EditableTapestryViewModel>
     )
   }
 
-  protected canApply(_patches: Patch[]): boolean {
+  protected canApply(patches: Patch[]): boolean {
```
Sachanski, line 51: Can you add some short comment describing what conditions are checked in order to determine whether a set of patches can be applied?

Also is the TODO still valid?

client/src/hooks/use-convert-to-pdf.ts
```diff
@@ -0,0 +1,13 @@
+import { useAsyncAction } from 'tapestry-core-client/src/components/lib/hooks/use-async-action'
+import { resource } from '../services/rest-resources'
+
+export function useConvertToPDF() {
```
Sachanski: I noticed that in the two places we are using this we are rendering the same toolbar elements. If you think it makes sense we could pass the item id to the hook and have it (the hook) return the respective `ItemToolbarMenuItem`, thus reducing some copy-paste?

Sachanski (dismissed):

client/src/pages/tapestry/view-model/undo-stack/tapestry-undo-stack.ts
zmarinov-astea: This checks solely if the patches are being applied to elements which exist, or will be created by the patches, in other words are not deleted by the server updates. 
I left the TODO there because canApply invalidates the whole array of patches, and we might want to keep those, which do not cause conflict...

Sachanski (approved):

PR #77, opened 2026-06-24
Add IIIF item type

dbvisel:
Adds support for **IIIF** ([International Image Interoperability Framework](https://iiif.io/)) images as a first-class item type, with deep-zoom playback via [OpenSeadragon](https://openseadragon.github.io/).

### What this does
- **Core:** IIIF manifest / Image API service resolution (`core/src/iiif.ts`) and derivation of IIIF manifest URLs from Internet Archive items.
- **Client / core-client:** IIIF item rendering and an OpenSeadragon-based viewer, plus item-factory, media, and export wiring.
- **Server:** resolves an IIIF item's source (an IA item URL or a direct manifest URL) to a IIIF Image API service endpoint on create/update.
- **Schema:** new `iiif` `ItemType` value and `Item.imageService` column + migration.
- **Shared:** IIIF item DTO/schema additions.
- **Dependency:** adds `openseadragon` (+ `@types/openseadragon`).

### Testing
- `tsc --build` passes for `shared`, `core`, `core-client`, `server`, and `client`.
- Manually verified IIIF playback in the app.


core/src/data-format/export/index.ts
```diff
@@ -213,11 +214,21 @@ class ParserV6 extends ExportParser<ExportV7> {
...
   public readonly version = 7
 
-  protected parseInternal(tapestry: ExportV7): ExportV7 {
+  protected parseInternal(tapestry: ExportV7): ExportV8 {
+    // The "iiif" item type is additive, so no item-level migration is required - just bump the version.
+    return { ...tapestry, version: 8 }
+  }
+}
+
+class ParserV8 extends ExportParser<ExportV8> {
```
zmarinov-astea, line 227: ParserV8 and ExportV8 are unnecessary

server/prisma/schema.prisma
```diff
@@ -188,6 +189,7 @@ model Item {
   height                       Int
   type                         ItemType
   webpageType                  WebpageType?
+  imageService                 String?
```
zmarinov-astea, line 192: Do we need the imageService column? The source attribute of iiif items is never used, except on item creation, so I think it's safe to save the imageSerivce as source. Can we fetch the manifest from the imageService? If we absolutely need the manifest, we can keep it and fetch the imageService every time we need it without storing it in the database. Haven't researched much, so you decide if we want to keep the manifest or the imageService itself as source.

core/src/iiif.ts
```diff
@@ -0,0 +1,157 @@
...
+function asString(value: unknown): string | undefined {
+  return typeof value === 'string' ? value : undefined
+}
+
+/** Returns the first element of an array, or the value itself if it isn't an array. */
+function first(value: unknown): unknown {
+  return Array.isArray(value) ? value[0] : value
+}
+
+function toNumber(value: unknown): number {
```
zmarinov-astea, line 45: Instead of all these helpers, can we use something simpler, or something we already have in the code, or a zod schema...?

server/src/resources/items.ts
```diff
@@ -99,6 +105,29 @@ async function resolveWebSource(item: ItemCreateDto | ItemUpdateDto) {
   item.webpageType = webSourceParser.webpageType
 }
 
+/**
+ * Normalizes an IIIF item's source. The client usually resolves the manifest itself (and sets
+ * `skipSourceResolution`), but when an item is created directly via the API we accept either an Internet
+ * Archive item URL (from which we derive the manifest) or a direct manifest URL, and resolve the IIIF
+ * Image API service endpoint from it if one wasn't provided.
+ */
+async function resolveIiifSource(item: (ItemCreateDto | ItemUpdateDto) & { type: 'iiif' }) {
```
zmarinov-astea, line 114: Do we need this source parsing functionality at the server?

server/prisma/schema.prisma
dbvisel: I understand what you're saying here, though I think there's an argument to be made for keeping this. Right now, we're letting people link to a `manifest.json` file, but there are other ways that people link to IIIF objects. If you look at the [IIIF Content State API](https://iiif.io/api/content-state/1.0/), for example, there are ways of linking to specific parts of an IIIF through base-64 encoded JSON in the initial URL that's passed. I don't think we need to deal with that right now, but that seems like something we are likely to want in the future.

zmarinov-astea: Hm, I don't think that it is a problem to put the JSON as a source or upload it to S3... What I am essentially saying is, that we shouldn't create redundancy in the database when one can be derived from the other.

dbvisel:
This branch is now current with `main` (merged, not rebased — kept the original commit intact since it already had review comments anchored to it) and addresses the review feedback below. Replying inline on each thread too, but summarizing the actual change here since the pieces fit together.

**`imageService` column removed.** An iiif item's `source` is the manifest URL. It now also carries the resolved IIIF Image API service endpoint, as a query param on that same URL (`core/src/iiif.ts`'s `withResolvedImageService`/`getResolvedImageService`/`getManifestUrl`). This removes the `Item.imageService` column and its migration, the dedicated schema field, and the dedicated transformer/zip-import code paths. `source` stays a real, fetchable manifest URL throughout — including the Content State API use case raised earlier in this thread.

**`ExportV8`/`ParserV8` removed.** V8 existed only because V7's own frozen item schema did not include the `iiif` type — `ExportV8Schema` was otherwise byte-identical to `ExportV7Schema`. Fixed by adding `IiifItemSchemaV7` to V7's schema directly. It is identical to `ImageItemSchemaV7`, minus the click-action fields, now that `imageService` needs no special casing. The `iiif` type ships as part of V7. No version bump needed.

**The `core/src/iiif.ts` helpers stay as they are.** Checked for an existing generic JSON-accessor utility to consolidate against — none exists in this codebase. A Zod schema would not remove the manifest-navigation logic itself, only relocate it, given how much the 2.x/3.x shapes genuinely differ. Verified the current parsing against real, live 2.x and 3.x manifests, including one with no image service at all (correctly returns `null`).

**`resolveIiifSource` (the server-side fallback) stays, but is smaller now.** It resolves and encodes one field instead of two. The real need behind it — direct item creation via the API without a browser to resolve client-side first (zip import, a script) — is still there; didn't find a way to drop it without losing that.

**Also done, since this PR predates it**: every comment this PR introduces is rewritten in ASD-STE100 (Simplified Technical English) style, per this repo's review conventions.

**Verified**: `npm run lint` clean across every workspace. The edited migration applies cleanly against a fresh local Postgres. The manifest fetch/encode/decode round-trip is verified against real live manifests (Wellcome Collection 2.x, the IIIF Cookbook's 3.x image-service example, and a real Internet Archive image item) via a standalone script, not just type-checked. Currently running locally for manual click-through testing.

core/src/data-format/export/index.ts
dbvisel: Removed ExportV8/ParserV8 — see the summary comment on the PR for the full explanation.

server/prisma/schema.prisma
dbvisel: Removed the imageService column — see the summary comment on the PR for the full explanation.

core/src/iiif.ts
dbvisel: Kept these as-is — see the summary comment on the PR for why.

server/src/resources/items.ts
dbvisel: Kept this, but it's smaller now — see the summary comment on the PR for the full explanation.

core/src/iiif.ts
```diff
@@ -0,0 +1,192 @@
...
+) {
+  return `${imageService.replace(/\/$/, '')}/${region}/${size}/${rotation}/${quality}.${format}`
+}
+
+/**
+ * Parse a IIIF Presentation API manifest, 2.x or 3.x. Extract the information needed to
+ * render its first canvas. Return `null` if the value is not a recognizable IIIF
+ * manifest, or has no image-bearing canvas.
+ */
+export function parseIIIFManifest(manifest: unknown): IIIFCanvas | null {
```
zmarinov-astea: Can we please use zod for parsing JSON

dbvisel:
closing this because it was superseded.

PR #78, opened 2026-07-02
Implement item layers

zmarinov-astea:
1. Added a layer column for the Items.
2. Every item strives to have a unique non-negative layer. Items with equal layers should be avoided, since their order is not determined, but they don't break the layering logic.
3. Items in pixi use zIndex now, since changing the snapshot resolution redraws the item on top otherwise.
4. Layers are normalized on the client every time the user clicks "Bring to top" or "Bring to back".
5. Refactored export schemas, so the V7 export schema can now handle the layers.


client/src/model/data/utils.ts
```diff
@@ -309,7 +313,8 @@ export async function createMediaItem<T extends MediaItemType>(
     dropShadow: true,
     position: ORIGIN,
     tapestryId,
-  } as MediaItemCreateDto & { type: T }
+    layer: DEFAULT_LAYER,
+  } satisfies MediaItemCreateDto as MediaItemCreateDto & { type: T }
```
Sachanski, line 317: Why the `satisfies` operator?

client/src/pages/tapestry/view-model/store-commands/items.ts
```diff
@@ -353,3 +356,21 @@ export function removeFromGroup(itemId: string): StoreMutationCommand<EditableTa
...
+): StoreMutationCommand<EditableTapestryViewModel> {
+  return (_, { store }) => {
+    const itemIds = new Set(ensureArray(ids))
+
+    const [frontItems, backItems] = partition(idMapToArray(store.get(`items`)), (i) =>
+      to === 'front' ? itemIds.has(i.dto.id) : !itemIds.has(i.dto.id),
+    )
+
+    sortBy(frontItems, ['layer'])
+    sortBy(backItems, ['layer'])
```
Sachanski: Doesn't `sortBy` return a new array and not sort it in-place? Moreover it might be a bit cleaner to sort the array returned by `idMapToArray(store.get('items'))`

client/src/pages/tapestry/view-model/utils.ts
```diff
@@ -181,3 +186,9 @@ export async function insertDataTransfer(
     })
   }
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function getMaxLayer(store: Store<EditableTapestryViewModel, any>) {
+  const elements = idMapToArray(store.get('items')).map((element) => element.dto.layer)
+  return elements.length === 0 ? DEFAULT_LAYER : Math.max(...elements)
```
Sachanski: You could probably also use `maxBy(idMapToArray(store.get('items')), 'dto.layer')?.dto.layer ?? DEFAULT_LAYER`

client/src/stage/controller/editor-item-controller.ts
```diff
@@ -315,28 +307,11 @@ export class EditorItemController extends ItemController {
...
-    const point = this.stage.pixi.tapestry.app.stage.worldTransform.applyInverse(cursorLocation)
-    this.editorStore.dispatch(
-      setSelectionRect(
-        new Rectangle(pointerSelection.rect.position, {
-          width: point.x - pointerSelection.rect.position.x,
-          height: point.y - pointerSelection.rect.position.y,
-        }),
-      ),
-    )
-  }
```
Sachanski, line 338: Why is this (and isDraggingItems) no longer needed?

core-client/src/stage/renderer/index.ts
```diff
@@ -89,7 +90,9 @@ export abstract class TapestryRenderer<
     const interactiveElement = this.store.get('interactiveElement')
     this.getGroups().forEach((group) => this.renderViewModel(group, selection, interactiveElement))
     this.getRels().forEach((rel) => this.renderViewModel(rel, selection, interactiveElement))
-    this.getItems().forEach((item) => this.renderViewModel(item, selection, interactiveElement))
+    sortBy(this.getItems(), 'dto.layer').forEach((item, zIndex) =>
```
Sachanski: I don't know how likely is this to cause performance issue, but isn't there an easy way to make sure we keep the items sorted and not sort them on every render?

client/src/pages/tapestry/view-model/store-commands/items.ts
```diff
@@ -353,3 +356,21 @@ export function removeFromGroup(itemId: string): StoreMutationCommand<EditableTa
...
+    const allItems = [...backItems, ...frontItems]
+    store.dispatch(...allItems.map((item, layer) => updateItem(item.dto.id, { dto: { layer } })))
```
Sachanski, line 372: This will make a single batch-mutation request instead of N updates, right?

client/src/model/data/utils.ts
zmarinov-astea: To enforce type constraint on the object. Otherwise a missing layer prop would not result a compile time error.

client/src/pages/tapestry/view-model/store-commands/items.ts
zmarinov-astea: Hm, sortBy really is not used properly. But I won't sort the initial array, since it is not guaranteed that lodash' partition will preserve the ordering.

client/src/stage/controller/editor-item-controller.ts
zmarinov-astea: This logic is being handled in the core-client item controller, and this function here was not being called at all. Also in the editor-item-controller the drag handler is set so that it is impossible to start a drag operation without dragging items.

core-client/src/stage/renderer/index.ts
zmarinov-astea: Yes, sorting is only managed by the zIndex, I put this additional sortBy to normalize the values, but we probably don't need it

client/src/pages/tapestry/view-model/store-commands/items.ts
zmarinov-astea: Yes, this updateItem command is also used when manipulating a selection - deleting, arranging, grouping....

Sachanski (approved):

PR #81, opened 2026-07-06
Animate tapestry actions

zmarinov-astea:
1. Animation settings can now be passed to navigation state.
2. Removed presentation shortcuts from item toolbars since we have a separate hook for them.
3. Fixed the viewer keyboard arrow pan, by tweaking the presentation hook.
4. Fixed a bug where ctrl key was being checked instead of meta key when dragging and resizing.
5. Added presentation navigation controls to shortcuts help pane.

### DIFF
--- a/client/src/stage/controller/editor-item-controller.ts
+++ b/client/src/stage/controller/editor-item-controller.ts
@@ -3,7 +3,10 @@ import { createEventRegistry } from 'tapestry-core-client/src/lib/events/event-r
 import { EventTypes } from 'tapestry-core-client/src/lib/events/typed-events'
 import { isMeta } from 'tapestry-core-client/src/lib/keyboard-event'
 import { TapestryStage } from 'tapestry-core-client/src/stage'
-import { ItemController } from 'tapestry-core-client/src/stage/controller/item-controller'
+import {
+  InternalNavigationState,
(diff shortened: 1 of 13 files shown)


client/src/stage/controller/editor-item-controller.ts
```diff
@@ -339,9 +345,10 @@ export class EditorItemController extends ItemController {
 
   private onDragItems(event: DragEvent<HoveredDragTarget>) {
     const { worldTransform } = this.stage.pixi.tapestry.app.stage
-    const guidelineSpacing = event.detail.originalEvent?.ctrlKey
-      ? null
-      : this.editorStore.get('viewportGuidelines.spacing')
+    const guidelineSpacing =
+      event.detail.originalEvent && isMeta(event.detail.originalEvent)
```
Sachanski, line 349: Why not keep using ctrl?

Sachanski (approved):

client/src/stage/controller/editor-item-controller.ts
zmarinov-astea: Ctrl is used on mac only for zooming. For every other shortcut we use the command key, and also this is how it is written in the help pane.

PR #82, opened 2026-07-08
Fix signaller connection initialization

zmarinov-astea:
Also refactored collaborator cursors, so they are always the same size and are animated when panning the viewport.

### DIFF
--- a/client/src/pages/tapestry/view-model/socket-manager.ts
+++ b/client/src/pages/tapestry/view-model/socket-manager.ts
@@ -77,9 +77,11 @@ export class SocketManager extends TypedEventTarget<TapestryUpdated | RTCSignall
   }
 
   activateSignaller() {
-    this.isSignallerActivated = true
-    if (this.connected) {
-      this.subscribeToSignalingService()
+    if (!this.isSignallerActivated) {
+      this.isSignallerActivated = true
+      if (this.connected) {
+        this.subscribeToSignalingService()
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


client/src/pages/tapestry/view-model/socket-manager.ts
```diff
@@ -48,7 +48,6 @@ export class SocketManager extends TypedEventTarget<TapestryUpdated | RTCSignall
       'rtc-signaling-message',
       this.tapestryId,
       (request: RTCSignalingMessage) => {
-        this.isSignallerActivated = true
```
Sachanski, line 51: Was this moved because of some race condition?

Sachanski (dismissed):

client/src/pages/tapestry/view-model/socket-manager.ts
zmarinov-astea: Yes, sometimes the socket reconnects initially several times before a response is received and the flag was not set. The flag's purpose is to reestablish rtc subscription after socket reconnections (for example switching viewer/edit mode), so that the socket is always subscribed to rtc-signaling-message channel after calling activateSignaller once. 
I can remove the isSignallerActivated flag altogether, although I want the signaling logic to be decoupled from the SocketManager itself

client/src/pages/tapestry/view-model/socket-manager.ts
```diff
@@ -78,7 +77,8 @@ export class SocketManager extends TypedEventTarget<TapestryUpdated | RTCSignall
   }
 
   activateSignaller() {
-    if (!this.isSignallerActivated) {
+    this.isSignallerActivated = true
+    if (this.connected) {
```
zmarinov-astea: It should be possible to call this only once here

Sachanski (approved):

PR #83, opened 2026-07-14
Add puppeteer autoconsent

zmarinov-astea:
[https://github.com/duckduckgo/autoconsent/blob/main/docs/puppeteer.md](https://github.com/duckduckgo/autoconsent/blob/main/docs/puppeteer.md)
Removes cookie banners in around 40% of the cases.
Currently enabled only for PDF conversion.
Moved PDF button to more menu since it is not so important.
Set a minimum size for PDFs since its layout is bad when its width is too small, and also too many pages appear if the height is too small.

### DIFF
--- a/server/src/tasks/autoconsent.ts
+++ b/server/src/tasks/autoconsent.ts
@@ -1,0 +1,131 @@
+import { fileURLToPath } from 'node:url'
+import { readFileSync } from 'node:fs'
+import { dirname, resolve } from 'node:path'
+import { Page } from 'puppeteer'
+import { pageEval } from './utils'
+import {
+  BackgroundMessage,
+  Config,
+  ContentScriptMessage,
+  RuleBundle,
+} from '@duckduckgo/autoconsent'
+
+export class AutoconsentError extends Error {}
+
+export const autoconsentScript = readFileSync(
+  resolve(
+    dirname(fileURLToPath(import.meta.resolve('@duckduckgo/autoconsent'))),
+    'autoconsent.playwright.js',
+  ),
+  'utf8',
+)
+
+// Load the rule bundles shipped with the package
+const rules = JSON.parse(
+  readFileSync(
+    fileURLToPath(import.meta.resolve('@duckduckgo/autoconsent/rules/rules.json')),
+    'utf8',
+  ),
+) as RuleBundle
+
+// See https://github.com/duckduckgo/autoconsent/blob/main/docs/api.md
+const autoconsentConfig: Partial<Config> = {
+  enabled: true,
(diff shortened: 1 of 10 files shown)


server/src/tasks/convert-to-pdf.ts
```diff
@@ -74,6 +80,9 @@ export async function convertToPdf({ itemId }: JobTypeMap['convert-to-pdf']) {
       deletedIds: { items: [item.id] },
     })
   } catch (error) {
+    if (s3Key) {
+      await s3Service.tryDeleteObject(s3Key)
+    }
```
Sachanski: Do we really need that? Won't it get picked up by the cron job that cleans up the s3 bucket?

server/src/tasks/autoconsent.ts
```diff
@@ -0,0 +1,131 @@
...
+    messages: false,
+    waits: false,
+  },
+}
+
+const sendMessage = async (page: Page, message: BackgroundMessage) =>
+  pageEval(
+    page,
+    (w, msg) => {
+      const window = w as { autoconsentReceiveMessage?: (msg: BackgroundMessage) => Promise<void> }
```
Sachanski: I guess we could extract this as a type/interface (type AutoconsentWindow = `{ ... }`) and simplify the expression to `window.autoconsentReceiveMessage?.(msg)`

server/src/tasks/autoconsent.ts
```diff
@@ -0,0 +1,131 @@
...
+  )
+
+export async function attatchAutoconsent(page: Page, timeout: number): Promise<string> {
+  return new Promise((res, rej) => {
+    let completed = false
+
+    function complete(msg: string, resolve: true): void
+    function complete(msg: unknown, resolve: false): void
+    function complete(msg: unknown, resolve: boolean) {
+      if (!completed) {
```
Sachanski: Might be a bit cleaner to reverse the condition so the entire function logic is not indented

server/src/tasks/autoconsent.ts
```diff
@@ -0,0 +1,131 @@
...
+
+          case 'autoconsentDone':
+            complete(`Duration: ${message.duration}ms`, true)
+            break
+          case 'autoconsentError':
+            complete(message.details, false)
+            break
+          case 'report':
+            if (message.state.lifecycle === 'nothingDetected') {
+              complete('Nothing detected', false)
```
Sachanski: Should this reject the promise?

Sachanski (dismissed):

Sachanski (approved):

PR #86, opened 2026-08-03
Double click on canvas

Sachanski:
* The gesture detector now detects a double-click event
* When the user double-clicks on an item/rel/selection/group the respective entity is fit on the screen (without being activated)

* Changed the wording for bringing an item to the bottom of the z-axis
* Added a toast after the user moves an item on the z-axis
* Added some margin to the converted pdf pages


client/src/components/tapestry-elements/item-toolbar/index.tsx
```diff
@@ -257,6 +257,10 @@ export function useEditMoreMenu({
         dtoArray.map(({ id }) => id),
         to,
       ),
+      setSnackbar({
+        text: 'Z-order updated (deselect the item to see the change)',
```
zmarinov-astea: Z-order could sound unfamiliar to some users, maybe more appropriate would be "Item sent to back".

core-client/src/view-model/utils.ts
```diff
@@ -165,13 +165,16 @@ export function itemsFocusRect<I extends ItemViewModel>(
...
+
+export function zoomToPoint(tapestry: TapestryViewModel, step: number, point: Point) {
   const {
-    size: { height, width },
     transform: { scale, translation },
   } = tapestry.viewport
-  const center = { x: width / 2, y: height / 2 }
   const minScale = getMinScale(tapestry.viewport, idMapToArray(tapestry.items))
-  return scaleBy(scale, translation, step, center, minScale, MAX_SCALE)
+  return scaleBy(scale, translation, step, point, minScale, MAX_SCALE)
```
zmarinov-astea: Here we don't check the canvas limits, and by click-zooming on the  edges of the canvas we can drift arbitrarily far away from the center.

core-client/src/stage/controller/item-controller.ts
```diff
@@ -94,6 +108,37 @@ export abstract class ItemController implements TapestryStageController {
...
+
+    switch (hoverTarget.type) {
+      case 'item':
+        return this.store.dispatch(focusItems([hoverTarget.modelId]))
+      case 'rel':
+        return this.store.dispatch(focusRel(hoverTarget.modelId))
+      case 'group':
+        return this.store.dispatch(focusGroup(hoverTarget.modelId))
+      case 'multiselection':
+        return this.store.dispatch(focusMultiselection())
```
zmarinov-astea, line 138: Focusing a multiselection doesn't seem to work in viewer mode. Also I think we need to have addToolbarPadding: true when focusing multiselection

core-client/src/view-model/store-commands/viewport.ts
```diff
@@ -305,6 +305,26 @@ export function focusGroup(
...
+    if (!rel) return
+
+    store.dispatch(focusItems([rel.dto.from.itemId, rel.dto.to.itemId]))
+  }
+}
+
+export function focusMultiselection(): StoreMutationCommand<TapestryViewModel> {
+  return (_, { store }) => {
+    const { itemIds, groupIds } = store.get('selection', ['itemIds', 'groupIds'])
+    const items = idMapToArray(store.get('items'))
```
zmarinov-astea: We have getSelectionItems function for this.

client/src/components/tapestry-elements/item-toolbar/index.tsx
strendafilov: If we want to shorten the text, I think that "Layer updated" is also a good choice.

core-client/src/stage/gesture-detector.ts
```diff
@@ -245,10 +247,23 @@ export class GestureDetector extends TypedEventTarget<
...
+          originalEvent: event,
+        })
+      } else {
+        this.firstClickTimeout = window.setTimeout(() => {
+          this.firstClickTimeout = undefined
+          this.dispatchEvent('click', {
+            hoverTarget: obtainHoverTarget(this.stage, event),
+            originalEvent: event,
+          })
+        }, 200)
```
strendafilov: I am not absolutely sure, but isn't a 200 ms range too short ( I have checked in google and the default timeout window to register a double-click on most operating systems like Windows and macOS is 500 ms, probably that is too much but for web apps it is something like 300-350 ms)

strendafilov (dismissed):

core-client/src/view-model/utils.ts
Sachanski: Turns out all zooms are pulling the viewport.

core-client/src/stage/gesture-detector.ts
Sachanski: I will bump it to 250

core-client/src/stage/controller/item-controller.ts
Sachanski: I have added the padding, but I think I will leave the view multiselection for a separate PR (maybe alongside the presentation navigation), since it might require a bit more changes than I would like to make

strendafilov (dismissed):

core-client/src/view-model/utils.ts
```diff
@@ -169,12 +170,31 @@ export function zoomToCenter(tapestry: TapestryViewModel, step: number) {
...
   const minScale = getMinScale(tapestry.viewport, idMapToArray(tapestry.items))
-  return scaleBy(scale, translation, step, point, minScale, MAX_SCALE)
+  const result = scaleBy(scale, translation, step, point, minScale, MAX_SCALE)
+
+  const [min, max] = getTranslationRange(
+    {
+      size: tapestry.viewport.size,
+      transform: {
+        scale: result.scale,
+        translation: mul(result.scale < 1 ? result.scale : 1 / result.scale, translation),
```
zmarinov-astea: For this to work I think we need to remove or make the translation property of getTranslationRange optional, since it exist only to prevent snapping when panning (i. e. if the viewport somehow went far away from the tapestry it doesn't clamp immediately after pan)

zmarinov-astea: I don't think that this snapping is a big problem since it can occur only in edit mode if the user is far away from the tapestry but there is a single item on the screen, and after he deletes it, the bounding rect shrinks and if the current translation is not accounted after pan/zoom he will get snapped to the bounding rect again...

Sachanski: I am not sure I understand what you are suggesting. Even in viewer mode if you go to top/right/bottom/left-most edge of the canvas and start zooming in you will be looking at a blank canvas.

zmarinov-astea: I am suggesting that in zoomToPoint we don't pass a translation property to getTranslationRange and in getTranslation range we return directly [minTranslation, maxTranslation] instead of [coordMin(minTranslation, translation), coordMax(maxTranslation, translation)] in this case

zmarinov-astea (approved):

PR #88, opened 2026-08-05
Limiting the canvas in view mode:

strendafilov:
In view mode:
Change maximum zoom out size - fit 75% of the content instead of 50%;
Lower the translation range;
Make scrollbars work properly with the new viewport sizes;
In edit mode:
Add some more padding in the translation range - contains 5% of the content, instead of 20px


client/src/pages/tapestry/view-model/store-commands/tapestry.ts
```diff
@@ -50,12 +56,30 @@ export function setInteractionMode(
...
+
+    const currentViewport = store.get('viewport')
+    const currentMode = store.get('interactionMode')
+
+    const isAlreadyUpdated =
+      currentMode === mode &&
+      currentViewport.minZoomContentRatio === minZoomContentRatio &&
+      currentViewport.maxOutsideRatio === maxOutsideRatio
+
+    if (isAlreadyUpdated) return
```
zmarinov-astea: Why don't we just use the old check if (mode === store.get('interactionMode')) return?
When opening the tapestry in edit mode, we should set the viewport parameters in client/src/model/data/utils.ts line 170 instead

core-client/src/view-model/index.ts
```diff
@@ -25,6 +29,8 @@ export interface Viewport {
   readonly lastUpdateTimestamp?: number
   readonly ready: boolean
   readonly isZoomingLocked?: boolean
+  readonly minZoomContentRatio: number
+  readonly maxOutsideRatio: number
```
zmarinov-astea: I think translationPaddingRatio sounds more appropriate, and we can set it to denote how much the center of the viewport can go outside the tapestry bounds. In this case the values would be 0.05 for viewer mode and 0.45 for edit mode. Maybe we should add a comment for that

core-client/src/view-model/index.ts
```diff
@@ -17,6 +17,10 @@ export const MIN_RESTRICTED_SCALE = 0.5
 export const MAX_RESTRICTED_SCALE = 1.5
 export const ZOOM_STEP = 0.1
 export const MAX_INITIAL_SCALE = 1
+export const MIN_ZOOM_RATIO_EDIT_MODE = 0.5
```
zmarinov-astea: We should remove these EDIT MODE constants from here since core-client 'does not know' about edit mode. Edit mode is an extension of the client project, so we should move them somewhere in the client folder, not in core-client

client/src/pages/tapestry/view-model/store-commands/tapestry.ts
```diff
@@ -50,12 +56,30 @@ export function setInteractionMode(
...
 
     store.dispatch(
       selectItem(null),
       (model) => {
         model.interactionMode = mode
+        model.viewport = {
+          ...model.viewport,
+          minZoomContentRatio,
```
zmarinov-astea: here we can write model.viewport.minZoomContentRatio = minZoomContentRatio instead.
Similar for maxOutsideRatio (translationPaddingRatio)

core-client/src/view-model/utils.ts
```diff
@@ -52,10 +54,17 @@ import { Easing } from '@tweenjs/tween.js'
...
 const CONTENT_FIT_PADDING = 16
+export const DEFAULT_VIEWPORT_LIMITS = {
+  minZoomContentRatio: MIN_ZOOM_RATIO_VIEW_MODE,
+  maxOutsideRatio: MAX_OUTSIDE_RATIO_VIEW_MODE,
+}
 
 export function viewModelFromTapestry(
   tapestry: Tapestry,
   presentationSteps: PresentationStep[],
+  viewportLimits: Partial<
```
zmarinov-astea: we can make this argument optional instead of giving it a default value, since we set the default values below either way.

core-client/src/view-model/utils.ts
```diff
@@ -224,35 +236,41 @@ export function getTheme(tapestry: TapestryViewModel) {
   return THEMES[tapestry.theme]
 }
 
-const MIN_ZOOM_CONTENT_RATIO = 0.5
-const MAX_MIN_SCALE = 0.5
+//This coefficient restricts the maximum zoom-out by requiring the content's box
+//to occupy at least 75% of the viewports dimensions
+const MAX_MIN_SCALE = 0.75
```
zmarinov-astea: This comment is inappropriate for MAX_MIN_SALE. Also I think we can set back MAX_MIN_SCALE to 0.5

core-client/src/view-model/utils.ts
```diff
@@ -224,35 +236,41 @@ export function getTheme(tapestry: TapestryViewModel) {
...
 
-const TRANSLATION_RANGE_PADDING: Vector = { dx: 20, dy: 20 }
 export function getTranslationRange(tapestry: TapestryViewModel): [Vector, Vector] {
   const {
     transform: { scale, translation },
     size: { width, height },
+    maxOutsideRatio,
   } = tapestry.viewport
+
+  const minVisibleRatio = 1 - maxOutsideRatio
```
zmarinov-astea: minVisibleRatio and maxOutsideRatio sound too confusing for me. If we use translationPaddingRatio we can just add mul(0.5 - translationPaddingRatio, viewportOffset) for minTranslation and mul(0.5 + translationPaddingRatio, viewportOffset) for maxTranslation. We use the center of the viewport as a reference point which is more clean in my opinion.

core-client/src/view-model/utils.ts
```diff
@@ -278,8 +296,8 @@ export function getScrollbarPositions<I extends ItemViewModel>(
     height: viewportSize.height / scale,
   }
   const canvasPadding: Vector = {
-    dx: viewportSizeOnTapestry.width - TRANSLATION_RANGE_PADDING.dx / scale,
-    dy: viewportSizeOnTapestry.height - TRANSLATION_RANGE_PADDING.dy / scale,
+    dx: viewportSizeOnTapestry.width * viewport.maxOutsideRatio,
```
zmarinov-astea: Here would be (0.5 + translationPaddingRatio)

core-client/src/view-model/index.ts
```diff
@@ -17,6 +17,10 @@ export const MIN_RESTRICTED_SCALE = 0.5
...
+export const MAX_OUTSIDE_RATIO_EDIT_MODE = 0.95
+export const MIN_ZOOM_RATIO_VIEW_MODE = 0.75
```
zmarinov-astea: Have you tried a value of 0.67 for MIN_ZOOM_RATIO_VIEW_MODE? Can't decide which is better...

client/src/pages/tapestry/view-model/store-commands/tapestry.ts
strendafilov: You are absolutely right. Initializing the viewport limits in fromTapestryDto() is much cleaner than patching the state here.

core-client/src/view-model/utils.ts
strendafilov: Yes and now I have found that i can set the default values in the viewport like this 
...DEFAULT_VIEWPORT_LIMITS,
...viewportLimits,
 which is actually very cool.

core-client/src/view-model/index.ts
strendafilov: Actually I have tried many values and 0.75 fits perfect ( for me of course ). If you want I can send you a screenshot to see the difference and decide which is better

core-client/src/view-model/index.ts
strendafilov: Ok, I will move them in client project

client/src/model/data/utils.ts
```diff
@@ -167,6 +171,17 @@ export function fromTapestryDto(
 ): EditableTapestryViewModel {
   const presentationStepViewModels = presentationSteps.map((dto) => ({ dto }))
 
+  const viewportLimits = {
```
zmarinov-astea: Here we can just write viewportLimits = mode === 'edit' ? EDIT_VIEWPORT_LIMITS : DEFAULT_VIEWPORT_LIMITS, or just pass that directly as an argument

core-client/src/view-model/index.ts
zmarinov-astea: We decided to use maxTranslationRatio

core-client/src/view-model/index.ts
zmarinov-astea: Ok, we'll see how it looks on dev

core-client/src/view-model/utils.ts
```diff
@@ -265,11 +258,11 @@ export function getTranslationRange(tapestry: TapestryViewModel): [Vector, Vecto
 
   const minTranslation: Vector = add(
     neg(mul(scale, bottomRightOffset)),
-    mul(minVisibleRatio, viewportOffset),
+    mul(minTanslationRatio, viewportOffset),
```
zmarinov-astea: I think we can remove minTranslationRatio variable and just paste 1 - maxTranslationRatio here since the ratio is actually the same, and we don't have min ratio. This 1 accounts for adjusting to the upper left corner of the viewport

core-client/src/view-model/index.ts
```diff
@@ -29,8 +25,16 @@ export interface Viewport {
...
+  //Minimum ratio of the content box that must remain visible inside the viewport when zoomed out.
+  //- View Mode: 0.75 (requires 75% of content visible)
+  //- Edit Mode: 0.50 (requires 50% of content visible)
   readonly minZoomContentRatio: number
-  readonly maxOutsideRatio: number
+
+  //Maximum fraction of the viewport dimensions allowed to overflow outside the content bounds when panning.
+  //Restricts how far the canvas can be translated relative to the screen.
+  //- View Mode: 0.55 (Max 55% off-canvas - guarantees at least 45% of the screen covers content)
+  //- Edit Mode: 0.95 (Max 95% off-canvas - provides maximum reach for elements near the edges)
```
zmarinov-astea: We should remove Edit and view mode from these comments since there are no modes in core-client at all. We could write instead "Example value: 0.55 (Max 55% off-canvas ..."

core-client/src/view-model/utils.ts
```diff
@@ -55,16 +53,14 @@ import { CONTINUOUS_ZOOM_SPEED } from './store-commands/viewport.js'
 
 const CONTENT_FIT_PADDING = 16
 export const DEFAULT_VIEWPORT_LIMITS = {
```
zmarinov-astea: We can also create a type here ViewportLimits, which we could apply to the EDIT_VIEWPORT_LIMITS as well to enforce typing. 
`type ViewportLimits = Pick<Viewport, 'minZoomContentRatio' | 'maxTranslationRatio'>`

client/src/pages/tapestry/view-model/store-commands/tapestry.ts
```diff
@@ -56,30 +59,23 @@ export function setInteractionMode(
...
     store.dispatch(
       selectItem(null),
       (model) => {
         model.interactionMode = mode
-        model.viewport = {
-          ...model.viewport,
-          minZoomContentRatio,
-          maxOutsideRatio,
-        }
+        model.viewport.minZoomContentRatio = minZoomContentRatio
```
zmarinov-astea: now that we have the two constants we can bring back 
`model.viewport = {
          ...model.viewport,
            ...mode === 'edit' ? EDIT_VIEWPORT_LIMITS : DEFAULT_VIEWPORT_LIMITS,
        }
`

core-client/src/view-model/index.ts
```diff
@@ -25,6 +25,16 @@ export interface Viewport {
   readonly lastUpdateTimestamp?: number
   readonly ready: boolean
   readonly isZoomingLocked?: boolean
+  //Minimum ratio of the content box that must remain visible inside the viewport when zoomed out.
```
zmarinov-astea: also this is not a ratio of the contet box, but a ratio of the viewport, which should be filled with content

core-client/src/view-model/utils.ts
strendafilov: Honestly, since it's only used in two places, I am not sure how much actually we need types for this, but let's say that is some kind of type safety.

zmarinov-astea (approved):

PR #90, opened 2026-08-06
Fix scrollbar disappear

### LINKED ISSUE #62: Scrollbars not disappearing
Scrollbars should disappear 500ms after viewport updates, but sometimes they stay because there is no state update.

strendafilov:


core-client/src/components/tapestry/viewport-scrollbars/index.tsx
```diff
@@ -1,21 +1,32 @@
...
   const { viewport, items } = useStoreData(['viewport', 'items'])
+  const [isVisible, setIsVisible] = useState(false)
+
   const scrollbarPositionsRef = useRef<ReturnType<typeof getScrollbarPositions>>({
     horizontal: { offset: 0, size: 0 },
     vertical: { offset: 0, size: 0 },
   })
 
-  const isVisible = Date.now() - (viewport.lastUpdateTimestamp ?? 0) < 500
+  useEffect(() => {
```
zmarinov-astea: This is OK, but since Rado suggests [avoiding using useEffect](https://react.dev/learn/you-might-not-need-an-effect) when possible, we can instead use a [debounced](https://lodash.com/docs/4.18.1#debounce) function:
`const hideToolbars = useRef(debounce(() => setIsVisible(false), SCROLLBAR_DISAPPEAR_DELAY))`  
Then we can add another state for remembering the lastUpdateTimestamp and  every time it changes just call setIsVisible(true) and call hideToolbars.current function. Also we can update the scrollbar positions only when the timestamp changes

Sachanski: We have a very similar situation in `client/src/components/collaborator-cursors/index.tsx`. We can maybe extract a reusable hook
```
export function useRecentlyChanged<T>(value: T, delay: number) {
  const [prev, setPrev] = useState(value)
  const [changed, setChanged] = useState(true)

  useEffect(() => {
    const timeout = window.setTimeout(() => setChanged(false), delay)
    return () => clearTimeout(timeout)
  }, [prev, delay])

  if (!isEqual(prev, value)) {
    setPrev(value)
    setChanged(true)
  }

  return changed
}
```
and use it in both places

Sachanski (dismissed):

Sachanski (approved):

PR #91, opened 2026-08-10
Move format icons to top right

zmarinov-astea:


core-client/src/lib/pixi.ts
```diff
@@ -97,3 +97,43 @@ export function drawDashedPolyline(graphics: Graphics, points: Point[], dash: nu
     }
   }
 }
+
+export function drawRoundedRect(
```
Sachanski, line 101: I find some similarities between this and `roundedRectPolyline` however I can't evaulate if they are enough to warrant some kind of code reuse.

Sachanski (approved):

core-client/src/lib/pixi.ts
zmarinov-astea: The only difference is that roundedRectPolyline calculates the arc points, whereas I use the arcTo command from pixi. Maybe the difference is in the performance since pixi calculates the number of steps relative to the absolute arc length `Math.max(6, Math.floor(6 * Math.pow(radius, 1 / 3) * (dist / Math.PI))))`, whereas roundedRectPolyline takes the current scale into account when using it for groups

Sachanski: Got it - I'd leave it azis for now.

zmarinov-astea: It really looks a little jagged on the highest zoom level, but we will need more complicated state management if we want to re-render it on different zoom levels. 
I will make a little tweak to the icon size calculation before merging.

PR #92, opened 2026-08-11
Duplicate image assets when forking

zmarinov-astea:
Also:
1. Fixed a bug where deleting an item did not delete its thumbnail image assets.
2. Added option to generate tapestry thumbnail job to generate thumbnails for all items.
3. Changed autoconsent files filename resolution to avoid compilation errors.

### DIFF
--- a/server/src/tasks/create-tapestry.ts
+++ b/server/src/tasks/create-tapestry.ts
@@ -1,10 +1,10 @@
-import { getCopyName, IdMap, isHTTPURL, mapIds } from 'tapestry-core/src/utils.js'
+import { getCopyName, IdMap, idMapToArray, isHTTPURL, mapIds } from 'tapestry-core/src/utils.js'
 import { JobTypeMap } from './index.js'
 import { prisma } from '../db.js'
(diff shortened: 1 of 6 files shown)


server/src/tasks/create-tapestry.ts
```diff
@@ -71,11 +71,15 @@ async function forkTapestry(job: TapestryCreateJob) {
           }
         }
 
+        const thumbnanilIds = compact(tapestry.items.map((i) => i.thumbnailId))
+        const thumbnailIdMap = fromPairs(thumbnanilIds.map((id) => [id, crypto.randomUUID()]))
```
Sachanski: I have the feeling that this is the fifth such id map, but it is not like I can suggest an abstraction that won't be attrocious

Sachanski (dismissed):

server/src/tasks/create-tapestry.ts
zmarinov-astea: Actually there are only two id maps with crypto.randomUUID in this file. I will try to rearrange them in the functions

Sachanski (approved):

PR #93, opened 2026-08-13
Add rich text editor for comments

strendafilov:


client/src/components/assign-action-button/index.tsx
```diff
@@ -20,44 +16,81 @@ type ActionItemDto = Extract<ItemDto, { type: ActionItemType }>
...
   initialAction?: string
+  initialText?: string
+  action?: string
+  onActionChange?: (value: string) => unknown
+  text?: string
+  onTextChange?: (value: string) => unknown
 }
 
-function AssignActionModal({
+export function AssignActionModal({
```
zmarinov-astea: I think it would be cleaner to create a separate AddLinkModal whose job would be to just get the user input, and which has an optional text field, then we can refactor the AssignActionModal to be a wrapper of AddLinkModal. We will use the AddLinkModal directly for the comments and the text boxes

client/src/components/assign-action-button/index.tsx
```diff
@@ -20,44 +16,81 @@ type ActionItemDto = Extract<ItemDto, { type: ActionItemType }>
...
-            icon="left_click"
-            aria-label="Attach items"
-            tooltip={{ side: 'bottom', children: 'Attach items' }}
-            onClick={() => onSelectItem()}
-          />
+          {showTextField && (
+            <Input
+              value={text}
+              onChange={(e) => setText(e.target.value)}
+              placeholder="Input link name"
```
zmarinov-astea: "Input text (optional)" would be less confusing than "Input link name" for a placeholder I think. Also we really need to make it optional

client/src/components/assign-action-button/index.tsx
```diff
@@ -20,44 +16,81 @@ type ActionItemDto = Extract<ItemDto, { type: ActionItemType }>
...
-            placeholder="Input a link or select an item"
-          />
-          <IconButton
-            icon="left_click"
-            aria-label="Attach items"
-            tooltip={{ side: 'bottom', children: 'Attach items' }}
-            onClick={() => onSelectItem()}
-          />
+          {showTextField && (
+            <Input
```
zmarinov-astea: We can also add labels "Text" and "Link" to the inputs as in slack

client/src/components/message-input/styles.module.css
zmarinov-astea, line 1: We might need to set a bigger width for the side pane, since the input stretches it. We can set it for the CommentsPane or change the SidePane general styles in some way... I think it will be better to apply the modification only when the comments pane is shown. Alternatively we can decrease the whole message input margin
Image

client/src/components/message-input/index.tsx
```diff
@@ -62,38 +136,87 @@ export function MessageInput({
...
+                onClose={closeLinkModal}
+                action={linkAction}
+                onActionChange={setLinkAction}
+                text={linkText}
+                onTextChange={setLinkText}
+                onApply={(url, text) => {
+                  const { action, actionType } = extractAction(url, tapestryPath, tapestryId)
+                  if (!action) return
+
+                  const href = actionType === 'internalLink' ? `${tapestryPath}?${action}` : action
```
zmarinov-astea: Why are we first extracting the action from the url and the appending the tapestryPath again? We can just write the query params to the href to indicate internal action

client/src/components/comments-list/comment/index.tsx
```diff
@@ -15,43 +15,76 @@ import { resource } from '../../../services/rest-resources'
...
+import { RichTextEditor } from 'tapestry-core-client/src/components/lib/rich-text-editor'
+import { useNavigate } from 'react-router'
+import { useTapestryPath } from '../../../hooks/use-tapestry-path'
 
 interface CommentProps {
   comment: CommentDto
   tapestryAuthorId: string
+  isEditing: boolean
+  onEditStart: () => void
+  onEditEnd: () => void
```
zmarinov-astea: instead of having onEditStart and onEditEnd we can just have setEditing: (value: boolean) => void

client/src/components/message-input/index.tsx
```diff
@@ -35,22 +47,84 @@ export function MessageInput({
...
+  const [linkAction, setLinkAction] = useState('')
+  const [linkText, setLinkText] = useState('')
+
+  const tapestryId = useTapestryData('id')
+  const tapestryPath = useTapestryPath('view')
+
+  const generateLink = useGenerateItemLink()
+  const items = useTapestryData('items')
+
+  const itemPicker = useItemPicker({
```
zmarinov-astea: We can try to put the itemPicker in the AddLinkModel since this could also be used for the text boxes

client/src/components/message-input/styles.module.css
```diff
@@ -8,36 +8,57 @@
...
+      border: 1px solid var(--theme-text-primary);
+    }
+
+    .editor-toolbar {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+      gap: 4px;
+      border-radius: 4px 4px 0 0;
+      padding: 4px 8px 4px 4px;
```
zmarinov-astea: I think we only need padding-right: 8px here

client/src/components/comments-list/comment/index.tsx
```diff
@@ -15,43 +15,76 @@ import { resource } from '../../../services/rest-resources'
...
-            </Tooltip>
-          </Text>
-        </>
-      ) : (
-        text
+      </Text>
+    )
+  }
+
+  const handleLinkClick = async (e: React.MouseEvent<HTMLDivElement>) => {
```
zmarinov-astea: In order to make the href work for forked and imported tapestries we can just save the query parameters when editing/creating a comment. Then here we will just check if the href is valid search param with URLSearchParams
`<a href="?focus=a123e-d124-asde-123e">Check out this item</a>`

client/src/components/message-input/index.tsx
```diff
@@ -35,22 +47,84 @@ export function MessageInput({
   const [joinPopup, setJoinPopup] = useState(false)
   const { user } = useSession()
 
+  const [selectionState, setSelectionState] = useState<SelectionState | undefined>(undefined)
+  const [isEditorReady, setIsEditorReady] = useState(false)
+  const editorApiRef = useRef<RichTextEditorApi | undefined>(undefined)
+
+  const [showModal, setShowModal] = useState(false)
+  const [linkAction, setLinkAction] = useState('')
+  const [linkText, setLinkText] = useState('')
```
zmarinov-astea: Can we combine the link text and link action into one state? Like { link: string, text: string }

client/src/components/comments-list/comment/index.tsx
zmarinov-astea: Can we always display the edit controls if we are currently editing?

client/src/components/message-input/styles.module.css
strendafilov: Removing just 4px of side padding is enough to solve this problem and it still looks good.

client/src/components/message-input/styles.module.css
strendafilov: 8 is too much padding, because the toolbar already has its own padding; this would result in a lot of space on the left but none on the right. Also, if we set the top and bottom padding to 8, it won't align with the input field below.

strendafilov: Never mind, I have read padding: 8px to all. You are right that we need only right padding, but in this case the action container is smaller than the input one

client/src/components/comments-list/comment/index.tsx
strendafilov: Actually we can and it will be better UX, because user will stop editing easier.

client/src/components/assign-action-button/index.tsx
strendafilov: Currently, both fields in the text box are mandatory and I’m actually wondering if it might be better to keep it that way, since you can insert a link right now without needing this modal.

client/src/components/comments-list/comment/index.tsx
strendafilov: Tried the URLSearchParams approach but it doesn't work reliably - URLSearchParams doesn't validate protocol/host, it just splits on & and =. So an external URL like https://google.com also parses as having "keys," meaning it'd get misclassified as internal too. Instead I'm marking the type explicitly at write time: internal hrefs get a ? prefix (query params only, no tapestryPath), external hrefs stay as full URLs. Then on click I just check href.startsWith('?') - deterministic, since we control the invariant at write time.

client/src/components/assign-action-button/index.tsx
zmarinov-astea: Yes, but we want the UI to be as user friendly as possible. Also the user might want to put some text but then decide to remove it but still add the link, so this will be helpful.

client/src/components/comments-list/comment/index.tsx
zmarinov-astea: Ok that's sufficient. One last thing - you are using the 'view' tapestry path, but collaborators with edit access can also leave comments, so I think we want to preserve the interaction mode after clicking and not always switching it to view mode

client/src/components/add-link-modal/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+export function AddLinkModal({
+  onClose,
+  onApply,
+  initialLink,
+  initialText,
+  showTextField = false,
+  link: controlledLink,
+  onLinkChange,
+  text: controlledText,
+  onTextChange,
```
zmarinov-astea: In which case do we want to set the link and text state externally? I don't think we need link, onLinkChange, text, onTextChange.

client/src/components/message-input/styles.module.css
```diff
@@ -4,40 +4,61 @@
...
     .message-input {
-      background-color: var(--theme-background-neutral);
       color: var(--theme-text-primary);
-      border-radius: 4px;
 
       &::placeholder {
         color: var(--theme-text-tertiary);
       }
+      /* stylelint-disable-next-line selector-class-pattern */
+      :global(.ProseMirror),
```
zmarinov-astea: I think we can use .tiptap instead of .ProseMirror here to comply with the lint rule

client/src/components/comments-list/comment/styles.module.css
```diff
@@ -49,4 +49,10 @@
       font-style: normal;
     }
   }
+  /* stylelint-disable-next-line selector-class-pattern */
+  :global(.ProseMirror),
```
zmarinov-astea: Isn't this style already applied in the message input component?

client/src/components/message-input/index.tsx
```diff
@@ -20,9 +30,10 @@ export interface MessageInputProps {
   endAdornment?: ReactNode
 }
 
+const TRAILING_EMPTY_PARAGRAPHS_REGEX = /(<p>(\s|<br\s*\/?>)*<\/p>)+$/gi
+
 export function MessageInput({
   onSubmit,
-  onPaste,
   disabled,
   placeholder,
```
zmarinov-astea, line 39: Why are we not setting the placeholder anywhere? It was "Add comment" before

client/src/components/comments-list/comment/index.tsx
```diff
@@ -15,43 +15,75 @@ import { resource } from '../../../services/rest-resources'
...
   onEdit?: (newComment: CommentDto) => void
 }
 
 function CommentContent({ createdAt, updatedAt, deletedAt, text }: CommentDto) {
   const lastUpdate = isEqual(updatedAt, createdAt) ? undefined : updatedAt
-  return (
-    <Text component="div" variant="bodySm" className={styles.commentText}>
-      {deletedAt ? (
+  const navigate = useNavigate()
+  const tapestryPath = useTapestryPath('view')
```
zmarinov-astea: can we pass the current interaction mode to useTapestryPath?

zmarinov-astea (dismissed):

client/src/components/comments-list/comment/styles.module.css
strendafilov: No, the styles aren't applied here because we have a separate field for the comment where the RTE styles are already applied.

zmarinov-astea: Ah, ok, I thought the html is rendered directly in the comment, not through RTE, which is also an option...?

zmarinov-astea: Also when selectors are separated by comma, this is a disjunction of selector, so p is being applied for all children of comment-text. To apply the style only if p is descendant from tiptap/ProseMirror write them consecutively without a comma

strendafilov: But I want to apply the style to the tiptap div also so removing the comma is not a full solution is it okay to use  

```
:global(.tiptap) {
    padding: 0 !important;
    margin: 0 !important;

    p {
      padding: 0 !important;
      margin: 0 !important;
    }
  }
```
`

zmarinov-astea: You can write 
```
:global(.tiptap) {
    &, p {
      padding: 0 !important;
      margin: 0 !important;
    }
  }
```

client/src/components/add-link-modal/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+import styles from './style.module.css'
+import { Id } from 'tapestry-core/src/data-format/schemas/common'
+
+export interface AddLinkModalProps {
+  onClose: () => unknown
+  onApply: (link: string, text?: string) => unknown
+  showTextField?: boolean
+  initialLink?: string
+  initialText?: string
+  link?: string
```
zmarinov-astea: Also we should remove   link?: 
  onLinkChange?:
  text?: 
  onTextChange?:
from the interface and make it non exported

client/src/components/add-link-modal/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+            disabled: !canApply,
+            onClick: () => onApply(link, text || undefined),
+          }}
+        >
+          <div className={styles.inputContainer}>
+            <div className={styles.actionContainer}>
+              {showTextField && (
+                <Input
+                  value={text}
+                  label={'Text'}
```
Sachanski: Here and below you can use string literals `label="Text"`

client/src/components/add-link-modal/style.module.css
```diff
@@ -7,22 +7,29 @@
...
+        padding-right: 50px;
       }
 
       > button {
         position: absolute;
-        right: 12px;
+        right: 10px;
         top: 50%;
-        transform: translateY(-50%);
+        transform: translateY(-15%);
```
Sachanski: Previously it was obvious that the combination of top: 50% and transform: translateY(-50%) was meant to center the element vertically. Now is not so clear what positioning we are aiming for. It might be a bit cleaner to add something like `endAdornment` propoerty to the input component, so the input itself aligns additional components to its actual typing area.

client/src/components/tapestry-elements/items/text/toolbar/index.tsx
```diff
@@ -23,7 +23,7 @@ const FOREGROUND_COLORS = OPAQUE_COLOR_PRESETS
 
 export type FormattingSubmenu = 'text-color' | 'text-background'
 
-interface TextItemToolbarOptions {
+export interface TextItemToolbarOptions {
```
Sachanski: Why is this exported?

client/src/components/comments-list/comment/index.tsx
```diff
@@ -15,43 +15,80 @@ import { resource } from '../../../services/rest-resources'
...
-            </Tooltip>
-          </Text>
-        </>
-      ) : (
-        text
+      </Text>
+    )
+  }
+
+  const handleLinkClick = async (e: React.MouseEvent<HTMLDivElement>) => {
```
Sachanski: You can pass this to the `RichTextEditor` component as (I would even consider inlining the logic)
```
<RichTextEditor
  events={{
    onClick: e => {
      ...
    }
  }}
/>
```
Also the name is a bit misleading - `handleLinkClick` assumes there is already a link that has been clicked. In truth the function first checks whether the clicked element is actually a link

core-client/src/components/lib/input/styles.module.css
```diff
@@ -1,4 +1,5 @@
 .root {
+  width: 100%;
```
Sachanski: Have you checked that this doesn't affect other instances of the `Input`?

core-client/src/components/lib/input/styles.module.css
```diff
@@ -17,6 +18,28 @@
...
+  }
+}
+
+.end-adornment {
+  position: absolute;
+  right: 8px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
```
Sachanski, line 29: The point was to remove the right padding and the absolute position.

I think the way to go here is to pass the root class to the new div that you have wrapped the input in and align the input and adornment inside it in a natural flex-row container. Also the whole `<div className={styles.endAdornment}` should be unnecessary

client/src/components/comments-list/comment/index.tsx
```diff
@@ -15,43 +15,84 @@ import { resource } from '../../../services/rest-resources'
...
+  if (deletedAt) {
+    return (
+      <Text component="div" variant="bodySm" className={styles.commentText}>
+        <em>[Deleted]</em>
+      </Text>
+    )
+  }
+
   return (
     <Text component="div" variant="bodySm" className={styles.commentText}>
```
Sachanski: Do we still need to be a `Text`?

client/src/components/message-input/index.tsx
```diff
@@ -1,12 +1,22 @@
+/* eslint-disable @typescript-eslint/no-empty-function */
```
Sachanski: Do not put global eslint-disable rules. That way we can accidentally ignore some warning that we can resolve.

You should remove this comment and use `noop` from lodash in the places where you have empty arrow functions.

client/src/components/message-input/index.tsx
```diff
@@ -62,38 +113,82 @@ export function MessageInput({
...
+              events={{
+                onCreate: () => setIsEditorReady(true),
+                onChange: setInput,
+                onSelectionChanged: setSelectionState,
+                onCreateLink: () => {
+                  handleCreateLink()
+                  return true
+                },
+                onKeyDown: (e) => {
+                  if (e.key === 'Enter' && (e.metaKey || e.ctrlKey)) {
```
Sachanski: Use `isMeta` instead.

client/src/components/message-input/index.tsx
```diff
@@ -35,22 +46,62 @@ export function MessageInput({
...
+      return
+    }
+    const selected = editorApiRef.current?.selectionText() ?? ''
+    setInitialLinkText(selected)
+    setShowModal(true)
+  }
+
+  const menuItems = textItemToolbar({
+    selection: selectionState,
+    tapestryId: '',
```
Sachanski: Does this have to be an empty string?

client/src/components/message-input/index.tsx
```diff
@@ -35,22 +46,62 @@ export function MessageInput({
...
+    editorAPI: editorApiRef,
+    itemBackgroundColor: null,
+    onBackgroundColorChange: () => {},
+    onColorChange: () => {},
+    onToggleMenu: () => {},
```
Sachanski: Here you can use lodash's `noop`.

Also you can leave a TODO here that we should change the types of the hook so that if a respective controls flag is false the property here cannot be passed, i.e. we passed controls.color === false, so it doesn't make sense to pass an `onColorChange` at all

client/src/components/message-input/index.tsx
```diff
@@ -35,22 +46,62 @@ export function MessageInput({
...
+    if (selectionState?.isLink) {
+      editorApiRef.current?.link('')
+      return
+    }
+    const selected = editorApiRef.current?.selectionText() ?? ''
+    setInitialLinkText(selected)
+    setShowModal(true)
+  }
+
+  const menuItems = textItemToolbar({
```
Sachanski: This should be renamed to something like `richTextEditorToolbar` since it is not used solely for text items anymore.

Sachanski (dismissed):

client/src/components/comments-list/comment/index.tsx
strendafilov: No, it can be div as well

client/src/components/message-input/index.tsx
strendafilov: Actually I can put it, but it is used only when the font size is changing. I have decided to set tapestryId to ' ' because we don't have fontSize in our RTE (actually the value of tapestryId doesn't matter at the moment) but I will add it in case that the font size is also added as a option in RTE.

Sachanski (approved):

PR #94, opened 2026-08-13
Fix dashboard cards horizontal order after auto reload

zmarinov-astea:
Also fixed a bug where the list auto scrolled to the top when the first element got loaded. This is useful for comments when new ones get added, but not for the tapestry list and the IA list

### DIFF
--- a/client/src/components/lazy-list/lazy-list-loader.ts
+++ b/client/src/components/lazy-list/lazy-list-loader.ts
@@ -97,9 +97,11 @@ export class LazyListLoader<T> extends Observable<ListResponseDto<T> & { state: 
   }
 
   private roundToSegment(value: number, round: 'up' | 'down', segmentLength: number) {
-    return round === 'up'
-      ? Math.trunc((value + segmentLength - 1) / segmentLength) * segmentLength
-      : Math.trunc(value / segmentLength) * segmentLength
+    return (
+      (Math.floor(value / segmentLength) +
+        (round === 'up' && value % segmentLength !== 0 ? 1 : 0)) *
+      segmentLength
+    )
   }
 
   /**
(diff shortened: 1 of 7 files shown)


client/src/pages/dashboard/tapestry-list/index.tsx
```diff
@@ -14,11 +14,41 @@ import { ListParamsInputDto } from 'tapestry-shared/src/data-transfer/resources/
...
+    }
+  }
+
+  const [columnCount, setColumnCount] = useState(determineColumnCount())
+
+  useEffect(() => {
+    const onWindowResize = () => setColumnCount(determineColumnCount())
+    window.addEventListener('resize', onWindowResize)
+    return () => window.removeEventListener('resize', onWindowResize)
+  }, [])
```
Sachanski, line 47: Can't we simply use `useResponsive` which returns the breakpoint when it changes?

client/src/components/lazy-list/lazy-list-loader.ts
```diff
@@ -96,8 +96,30 @@ export class LazyListLoader<T> extends Observable<ListResponseDto<T> & { state:
...
+  }
+
+  /**
+   * segmentLength splits the items into segments so that data.length and skip are always
+   * a multiple of segmentLength when possible.
+   * This is useful for grid like lists where a segment corresponds to one row
+   */
+  setRequestItems(requestItems: LazyListRequestItems<T>, segmentLength?: number) {
+    if (segmentLength) {
+      this.requestItems = async (skip: number, limit: number, signal: AbortSignal) => {
```
Sachanski: I think you can omit the types

Sachanski (dismissed):

client/src/pages/dashboard/tapestry-list/index.tsx
zmarinov-astea: Yes, but we need to update the state every time the column count changes, not when the breakpoint changes, and Breakpoint.XL can have many column counts.

Sachanski (approved):

PR #95, opened 2026-08-18
Add zoom-out after zooming an element  

strendafilov:


Sachanski (commented):
The approach with `requestAnimationFrame` seems a bit flaky. If for example the user starts panning immediatelly after double clicking the `transformsMatch` function will never return true and the recursion will go on forever.

Instead you can keep only the `currentTransform` as a property of the controller and pass it as an optional parameter to focusItems/focusRel/focusGroup/focusMultiselection.

From there (since focus Rel/Group/Multiselection call focusItems) inside `focusItems` you can get the result of `zoomToFit` and compare it against the passed argument and in case the linear transformations match you can call `transformViewport` with the passed argument instead of the target transfomration.

The `transformsMatch` function should probably go in some other file like `lib/geometry.ts` or `view-model/utils.ts`. Also its params should be of type `LinearTransformation` instead of `TapestryViewModel['viewport']['transform']`

As a side note you should change the cmd + double click condition to be `!hoverTarget && zoomOut`

core-client/src/view-model/utils.ts
```diff
@@ -427,3 +427,16 @@ export function getZoomParameters(fromScale: number, toScale: number, continuous
       : true,
   }
 }
+
+export function transformsMatch(
```
zmarinov-astea: Hm, I think we can avoid transformsMatch function if we set the translation to { dx, dy } if progres === 1 on line 150 in core-client/src/view-model/store-commands/viewport.ts Maybe some error arises from the calculations, we should try

core-client/src/stage/controller/item-controller.ts
```diff
@@ -114,10 +115,21 @@ export abstract class ItemController implements TapestryStageController {
     const zoomOut = isMeta(originalEvent)
     // If the user is holding the meta key down we will always zoom out,
     // irrespective of whether they are pointing at an item/rel/group/multiselection
-    if (!hoverTarget || zoomOut) {
+    if (!hoverTarget && zoomOut) {
```
Sachanski: I think I mislead you with my previous comment. The condition here should only be `if (!hoverTarget)`, since we want to be able to zoom in when double-clicking on an empty canvas, but want to zoom-out with ctrl/cmd _only_ if we have clicked on an empty canvas. You should also change the comment to reflect the up-to-date behaviour.

core-client/src/stage/controller/item-controller.ts
```diff
@@ -127,16 +139,30 @@ export abstract class ItemController implements TapestryStageController {
       return
     }
 
+    const currentTransform: LinearTransform = {
+      translation: { ...tapestry.viewport.transform.translation },
+      scale: tapestry.viewport.transform.scale,
+    }
```
Sachanski: We could shorten this a bit - `const currentTransform = structuredClone(tapestry.viewport.transform)`. The clone should not be needed, but it doesn't hurt to have it just in case

core-client/src/view-model/utils.ts
zmarinov-astea: An even more robust way would be to somehow compare the lastUpdateTimestamp of the viewport, this could solve many of our problems

Sachanski (dismissed):

core-client/src/stage/controller/item-controller.ts
```diff
@@ -127,16 +129,27 @@ export abstract class ItemController implements TapestryStageController {
       return
     }
 
+    const currentTransform = structuredClone(tapestry.viewport.transform)
+    const previousTransform = this.previousTransform ?? undefined
```
Sachanski: You can remove this coalesce and have
```
private previousTransform?: LinearTransform
```

core-client/src/stage/controller/item-controller.ts
```diff
@@ -127,16 +129,27 @@ export abstract class ItemController implements TapestryStageController {
...
+        break
       case 'group':
-        return this.store.dispatch(focusGroup(hoverTarget.modelId))
+        this.store.dispatch(focusGroup(hoverTarget.modelId, { previousTransform }))
+        break
       case 'multiselection':
-        return this.store.dispatch(focusMultiselection())
+        this.store.dispatch(focusMultiselection({ previousTransform }))
+        break
+      default:
```
Sachanski: This return should not be needed since the switch is exhaustive.

Sachanski (dismissed):

core-client/src/view-model/utils.ts
zmarinov-astea: Ok maybe there is no error if the animation is linear - the vaules are the same...

Sachanski (dismissed):

Sachanski (dismissed):

Sachanski (approved):

PR #96, opened 2026-08-18
Add Internet Archive search-query bulk import

dbvisel:
## Summary
- Recognizes Internet Archive search URLs (e.g. `https://archive.org/search?query=subject%3A%22Gondavalekar%22`) and opens the existing IA import picker so the user can choose which results to import, up to the shared `MAX_SELECTION` cap.
- Reuses the existing `HandleIAImportDialog`/`IAImport`/`import-items-list` picker mechanism (originally built for `IACollection`/`IAPlaylist`), adding a new `IASearchCollection` member and a sibling list component that queries the raw search string instead of `collection:<id>`.
- Imports by reference (embeds the archive.org item), consistent with how `IACollection`/`IAPlaylist` already behave.

## Test plan
- [x] `tsc --build` passes in `core` and `client`
- [x] `eslint` passes in `core` and `client`
- [x] `prettier --check` passes in `core` and `client`
- [ ] Manually paste an `archive.org/search?query=...` URL into a tapestry and confirm the picker opens, lists results, "select all" respects the 50-item cap, and confirmed items import correctly


client/src/components/handle-ia-import-dialog/import-items-list/search-list/index.tsx
zmarinov-astea, line 1: This looks like an almost complete copy of IACollectionList. Why don't we just remove IACollectionList and rename this to IASearchList instead. Then for collection we should just pass the query `collection:${collectionId}`, and the appropriate placeholder

client/src/components/handle-ia-import-dialog/import-items-list/search-list/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { useMemo, useState } from 'react'
+import { partial } from 'lodash-es'
+import { MAX_SELECTION } from '../..'
+import { LazyListLoader } from '../../../lazy-list/lazy-list-loader'
+import { useObservable } from 'tapestry-core-client/src/components/lib/hooks/use-observable'
+import { SelectAll } from '../select-all'
+
+function getSearchOpts(query: string) {
+  return {
+    q: query,
```
zmarinov-astea: Here we need to always exclude collections since we don't want them in the list `AND NOT mediatype:collection`

client/src/stage/item-factories.ts
```diff
@@ -158,6 +160,18 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
   }
 }
 
+const iaSearchCollectionFactory: ItemFactory = async (source, _mediaType, _tapestryId) => {
+  if (typeof source !== 'string' || !isHTTPURL(source)) return null
+
+  const parsed = parseIASearchURL(source)
+  if (!parsed) return null
+
+  const total = await fetchIASearchCount(parsed.query)
```
zmarinov-astea: here we should also exclude collections `AND NOT mediatype:collection`

client/src/stage/item-factories.ts
```diff
@@ -158,6 +160,18 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
   }
 }
 
+const iaSearchCollectionFactory: ItemFactory = async (source, _mediaType, _tapestryId) => {
```
zmarinov-astea: Can we use one IA factory which handles all IA logic? For example iaItemFactroy

core/src/internet-archive.ts
```diff
@@ -115,6 +115,27 @@ export async function iaAdvancedSearch<F extends IAAdvancedSearchFieldsQuery>(
   }
 }
 
+export function parseIASearchURL(source: unknown): { query: string } | null {
```
zmarinov-astea: Can we return just string or null here? Doing this we could rename it to parseIASearchURLQuery

client/src/stage/item-factories.ts
zmarinov-astea: Also this request might be a bit slow, but there is no other way to tell if the search result returns any items

core/src/internet-archive.ts
```diff
@@ -115,6 +115,37 @@ export async function iaAdvancedSearch<F extends IAAdvancedSearchFieldsQuery>(
...
+ * importable item - every IA search query this module builds (the raw search-URL case and the
+ * literal `collection:<id>` case alike) needs to exclude it, or the count and the list disagree
+ * about what's actually importable.
+ */
+export function excludeIACollections(query: string) {
+  return `(${query}) AND NOT mediatype:collection`
+}
+
+export async function fetchIASearchCount(query: string) {
+  const response = await iaAdvancedSearch({ q: excludeIACollections(query), pageSize: 1 })
```
zmarinov-astea: fetchIASearchCount should not depend on excludeIACollections, rather we should modify the argument before it is passed

client/src/stage/item-factories.ts
```diff
@@ -126,13 +128,29 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
   )
 }
 
-const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
+/**
+ * Handles every Internet Archive URL shape in one place: a search-results page
```
zmarinov-astea: We put comments only on core logic and complex math, we don't need a comment here

client/src/components/handle-ia-import-dialog/import-items-list/search-list/index.tsx
```diff
@@ -90,18 +91,22 @@ export async function requestCollectionItems(
   }
 }
 
-interface IACollectionListProps extends Omit<ImportItemsListProps, 'iaImport'> {
-  collectionId: string
+interface IASearchListProps extends Omit<ImportItemsListProps, 'iaImport'> {
+  query: string
+  /** Shown when the query returns no results - callers with a more specific case (e.g. "this
```
zmarinov-astea: This comment is not necessary

core/src/internet-archive.ts
```diff
@@ -115,6 +115,37 @@ export async function iaAdvancedSearch<F extends IAAdvancedSearchFieldsQuery>(
...
+    }
+    return query
+  } catch (error) {
+    console.warn(error)
+    return null
+  }
+}
+
+/**
+ * A "collection" is itself just another search result (`mediatype: collection`), never an
```
zmarinov-astea: This comment is also not necessary

core/src/internet-archive.ts
```diff
@@ -115,6 +115,37 @@ export async function iaAdvancedSearch<F extends IAAdvancedSearchFieldsQuery>(
   }
 }
 
+export function parseIASearchURLQuery(source: unknown): string | null {
+  if (typeof source !== 'string' || source === '') return null
+
+  try {
+    const url = new URL(source)
+    const query = url.searchParams.get('query')
```
zmarinov-astea, line 123: When searching the internet archive website I noticed that there is also a "tab" parameter in the url which indicates the opened tab where only specific item types are listed(books, text, radio...). Should we somehow add this type constraint to our query string?

core/src/internet-archive.ts
dbvisel: Fixed — `fetchIASearchCount` now just takes the query as-is; the one call site (`iaFactory`) passes `excludeIACollections(searchQuery)` explicitly. See 5b7b064.

client/src/stage/item-factories.ts
dbvisel: Removed.

client/src/components/handle-ia-import-dialog/import-items-list/search-list/index.tsx
dbvisel: Removed.

core/src/internet-archive.ts
dbvisel: Removed.

core/src/internet-archive.ts
dbvisel: Good catch — implemented. Checked the real site to see which tab produces which `tab=` value: most map directly onto our existing `IAMediaType` (tab=movies -> mediatype:movies, tab=etree -> mediatype:etree, tab=audio/software/image/data/texts likewise). `parseIASearchURLQuery` now folds that into the query as an added `mediatype:` clause when the tab is recognized, verified against the real advancedsearch.php API.

Deliberately not handling three cases, since none of them have a real `mediatype:` equivalent:
- `tab=collection` — a real mediatype value, but every query we build already excludes `mediatype:collection`, so honoring it would AND two contradictory clauses and always return zero results.
- `tab=radio` / `tab=tv` — not part of the mediatype facet at all (verified: no such IAMediaType value).
- `tab=fulltext` ("Text Contents") — that's full-text search mode, not a media-type restriction.

See 5b7b064.

zmarinov-astea: Since we don't support image types and full text search yet, I suggest we remove the tab parameter parsing functionality in this PR so we don't implement it partially.

dbvisel: Fixed — removed the `tab=` parsing entirely (the `TAB_TO_MEDIA_TYPE` map and the mediatype clause it added), so `parseIASearchURLQuery` just returns the raw `query` param again. See a5449f8.

zmarinov-astea (dismissed):

client/src/components/handle-ia-import-dialog/import-details/index.tsx
```diff
@@ -11,15 +11,36 @@ interface ImportDetailsProps {
...
+            </Text>
+            <Text variant={textVariant} lineClamp={2}>
+              {iaImport.total} results
+            </Text>
+          </div>
+        </div>
+        <Text variant={textVariant} component="div">
+          {iaImport.query}
+        </Text>
+      </div>
```
Sachanski, line 36: There is quite a bit of duplication between the JSX returned by this component. It would be a bit cleaner to extract some layout component that abstracts the common DOM containers, but since this is fairly contained I would be OK with leaveing it as it currently is.

At the very least let's add a TODO to potentially refactor this in the future

Sachanski (dismissed):

client/src/components/handle-ia-import-dialog/import-details/index.tsx
dbvisel: Added a TODO in b0b7258 - agreed this is worth a future shared-layout extraction, but leaving the two branches as-is for now given how contained they are.

zmarinov-astea (approved):

PR #97, opened 2026-08-21
Release/v1.2.0

Sachanski:


core-client/src/stage/gesture-detector.ts
```diff
@@ -245,10 +247,23 @@ export class GestureDetector extends TypedEventTarget<
...
+          originalEvent: event,
+        })
+      } else {
+        this.firstClickTimeout = window.setTimeout(() => {
+          this.firstClickTimeout = undefined
+          this.dispatchEvent('click', {
+            hoverTarget: obtainHoverTarget(this.stage, event),
+            originalEvent: event,
+          })
+        }, 250)
```
strendafilov, line 265: While using the dev environment, it often failed to register my double-clicks, so I’m wondering if it might be better to set it to 300 (though it’s not a huge deal).

strendafilov (approved):

zmarinov-astea:
We forgot to change the snackbar text to "Item brought to front/sent to back (deselect to see)"

core-client/src/stage/gesture-detector.ts
zmarinov-astea: AI: The default double-click timeout across major operating systems like Windows and most desktop environments is 500 milliseconds (0.5 seconds). This value defines the maximum time allowed between the first and second click for the system to register a double-click action.

zmarinov-astea: Yes but this way we will slow down the single click, which is not desirable

zmarinov-astea (approved):

core-client/src/stage/gesture-detector.ts
zmarinov-astea: AI: In cases where a single click and a double-click perform entirely conflicting actions, the ideal delay is between 200 and 300 milliseconds.Setting the delay to the full system standard of 500 milliseconds makes the interface feel sluggish and unresponsive to the user.

PR #98, opened 2026-08-21
Change set layer snackbar text

zmarinov-astea:


client/src/components/tapestry-elements/item-toolbar/index.tsx
```diff
@@ -258,7 +258,7 @@ export function useEditMoreMenu({
         to,
       ),
       setSnackbar({
-        text: 'Layer updated (deselect the item to see the change)',
+        text: `Item ${to === 'front' ? 'brought to front' : 'sent to back'} (deselect to see)`,
```
strendafilov: brought and sent should start with capital letters

strendafilov (approved):

PR #100, opened 2026-08-21
Preload item thumbnails

zmarinov-astea:
1. Low resolution thumbnails are now loaded in the TapestryLoader.
2. Added new thumbnailsInitialized prop which indicates when DOM elements should be rendered
3. Optimized loading by fetching items, rels, groups, and presentation steps concurrently, not sequentially.
4. Icon textures are now loaded in TapestryLoader and only once since they load too slow even [after a bugfix](https://github.com/pixijs/pixijs/issues/9531)
5. Fixed a bug where renditions were not unloaded after exiting the tapestry


client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
```diff
@@ -187,46 +187,65 @@ export class TapestryResourcesRepo extends ResourceRepo<TapestryResourceName, Ty
...
+        relsPromise ?? Promise.resolve(undefined),
+        groupsPromise ?? Promise.resolve(undefined),
+        presentationStepsPromise ?? Promise.resolve(undefined),
+      ])
+
+    result.tapestries = tapestriesResult && [tapestriesResult]
+    result.items = itemsResult
+    result.rels = relsResult
+    result.groups = groupsResult
+    result.presentationSteps = presentationStepsResult
```
Sachanski: We could probably do some TS magic to basically leave the code as it was but only remove the await's and only add a Promise.all before returning, but this is clear enough.

core-client/src/stage/controller/item-thumbnail-controller.ts
```diff
@@ -43,12 +50,29 @@ export class ItemThumbnailController implements TapestryStageController {
...
+      this.onInitialized()
+    } else {
+      this.fetchInitialThumbnails()
+    }
   }
 
   dispose(): void {
     this.store.unsubscribe(this.recalculateLOD)
     this.store.unsubscribe(this.onItemsChanged)
+    this.onItemsChanged({})
```
Sachanski, line 70: Why is this needed?

core-client/src/stage/controller/item-thumbnail-controller.ts
```diff
@@ -43,12 +50,29 @@ export class ItemThumbnailController implements TapestryStageController {
...
+    if (this.initialThumbnails) {
+      Object.entries(this.initialThumbnails).forEach(([itemId, rendition]) => {
+        if (!rendition) {
+          return
+        }
+        this.updateItemSnapshot(itemId, {
+          id: rendition.snapshotId,
+          texture: Texture.from(rendition.bitmap),
+        })
+        this.thumbnails[itemId] = { loadedRendition: rendition }
```
Sachanski: Not sure whether it is worth extracting some `set/applyThumbnail` function to avoid duplicating the call to updateItemSnapshot and setthing the rendition in the `thumbnails` map

Sachanski (dismissed):

core-client/src/stage/controller/item-thumbnail-controller.ts
zmarinov-astea: Before there was a bug that the renditions weren't unloaded after exiting the tapestry, so they accumulated in memory.

Sachanski (approved):

PR #101, opened 2026-08-24
Feature - add publisher's info in pdf captures

strendafilov:
Add additional information about the website publisher in the generated header/footer:
Header:
- Favicon of the page 
- Page title
- Date the pdf was generate

Footer:
- Page URL
- Pagination

Screenshot from 2026-08-24 12-14-51
Screenshot from 2026-08-24 12-15-53

### DIFF
--- a/server/src/tasks/convert-to-pdf.ts
+++ b/server/src/tasks/convert-to-pdf.ts
@@ -54,7 +54,34 @@ export async function convertToPdf({ itemId }: JobTypeMap['convert-to-pdf']) {
     const value = await convertWebpageToPdf(item.source, {
       width: Math.max(item.width, MIN_PDF_PAGE.width),
       height: Math.max(item.height, MIN_PDF_PAGE.height),
-      margin: { right: 20, left: 20 },
+      margin: { right: 20, left: 20, top: 60 },
+      displayHeaderFooter: true,
+      headerTemplate: `
+      <div style="
+        width: 100%;
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        background: #f5f6f8;
+        border: 1px solid #e2e4e8;
+        border-radius: 6px;
+        padding: 6px 12px;
+        margin: 0 30px 0; 
+        box-shadow: 0 1px 2px rgba(0,0,0,0.04);
+      ">
+            <span style="
+              font-size: 18px;
+              font-weight: 600;
+              color: #1a1a1a;
+            " class="title"></span>
+            <span style="
+              font-size: 12px;
+              color: #9a9da3;
+              margin-left: 12px;
+            " class="date"></span>
+          </div>
+      `,
+      footerTemplate: `<span></span>`,
     })
 
     const s3Key = tapestryKey(item.tapestryId, `${crypto.randomUUID()}.pdf`, true)
--- a/server/src/tasks/convert-to-pdf.ts
+++ b/server/src/tasks/convert-to-pdf.ts
@@ -1,7 +1,12 @@
 import { Page, PDFOptions } from 'puppeteer'
 import { JobTypeMap } from '.'
 import { prisma } from '../db'
-import { initWebpage, inNewBrowserPage, scheduleTapestryThumbnailGeneration } from './utils'
+import {
+  initWebpage,
+  inNewBrowserPage,
+  pageEval,
+  scheduleTapestryThumbnailGeneration,
+} from './utils'
 import { s3Service, tapestryKey } from '../services/s3-service'
 import { DBSubscriber } from '../socket'
 import { pick } from 'lodash-es'
@@ -22,24 +27,24 @@ const CONVERT_ITEM_PROPS = [
   'dropShadow',
 ] satisfies (keyof Item)[]
 
-async function getFaviconUrl(page: Page): Promise<string | null> {
-  return page.evaluate((): string | null => {
-    // The project doesn't include lib.dom in server's tsconfig, so
-    // we manually type only the browser API surface we need here.
-    const globalContext = globalThis as unknown as {
-      document: {
-        querySelector: (
-          selector: string,
-        ) => { getAttribute: (attr: string) => string | null } | null
-        baseURI: string
-      }
-    }
+type FaviconWindow = {
+  document: {
+    querySelector: (selector: string) => {
+      getAttribute: (attr: string) => string | null
+    } | null
+    baseURI: string
+  }
+}
 
-    const faviconHref = globalContext.document
+async function getFaviconUrl(page: Page): Promise<string | null> {
+  return pageEval(page, (window) => {
+    const browserWindow = window as FaviconWindow
+
+    const faviconHref = browserWindow.document
       .querySelector('link[rel="icon"], link[rel="shortcut icon"], link[rel="apple-touch-icon"]')
       ?.getAttribute('href')
 
-    return faviconHref ? new URL(faviconHref, globalContext.document.baseURI).href : null
+    return faviconHref ? new URL(faviconHref, browserWindow.document.baseURI).href : null
   })
 }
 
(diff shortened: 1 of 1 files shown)


server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +22,90 @@ const CONVERT_ITEM_PROPS = [
   'dropShadow',
 ] satisfies (keyof Item)[]
 
+async function getFaviconUrl(page: Page): Promise<string | null> {
+  return page.evaluate((): string | null => {
```
zmarinov-astea: We can use pageEval here instead from server/src/tasks/utils.ts Look at examples of how we are using it

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +22,90 @@ const CONVERT_ITEM_PROPS = [
...
+
+function buildHeaderTemplate(faviconDataUri: string | null): string {
+  return `
+    <div style="width: 100%; display: flex; justify-content: space-between; background: #f5f6f8; border: 1px solid #e2e4e8; padding: 6px 12px; margin: 0 30px 0;">
+      <div style="display: flex; align-items: center; gap: 8px; min-width: 0; max-width: 80%;">
+        ${faviconDataUri ? `<img src="${faviconDataUri}" style="width: 14px; height: 14px; border-radius: 2px; flex-shrink: 0;" />` : ''}
+        <span style="font-size: 12px; font-weight: 600; color: #1a1a1a; white-space: nowrap; overflow: hidden; text-overflow: ellipsis;" 
+            class="title"></span>
+      </div>
+      <span style="font-size: 10px; color: #9a9da3; margin-left: 12px; flex-shrink: 0;" class="date"></span>
```
zmarinov-astea: We don't need margin left here since we have justify content: space-between

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +22,90 @@ const CONVERT_ITEM_PROPS = [
...
+    return `data:${contentType};base64,${base64}`
+  } catch (error) {
+    console.error('>  Failed to fetch favicon', error instanceof Error ? error.message : error)
+    return null
+  }
+}
+
+function buildHeaderTemplate(faviconDataUri: string | null): string {
+  return `
+    <div style="width: 100%; display: flex; justify-content: space-between; background: #f5f6f8; border: 1px solid #e2e4e8; padding: 6px 12px; margin: 0 30px 0;">
```
zmarinov-astea: This container sems to overflow and is not centered on the page. You should put box-sizing: border-box to include the paddings

server/src/tasks/convert-to-pdf.ts
```diff
@@ -81,5 +157,6 @@ export async function convertToPdf({ itemId }: JobTypeMap['convert-to-pdf']) {
     })
   } catch (error) {
     console.error(`Error while converting item ${itemId} to pdf`, error)
+    throw error
```
zmarinov-astea: Don't throw the error here, this could stop the worker. Look at how errors are handled in other worker tasks

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +22,90 @@ const CONVERT_ITEM_PROPS = [
...
 async function convertWebpageToPdf(url: string, options?: PDFOptions) {
   let generator: ReturnType<typeof inNewBrowserPage<Uint8Array>> | undefined
   try {
     console.log(`Converting ${url} to pdf...`)
     generator = inNewBrowserPage(async function* (page, context): AsyncGenerator<Uint8Array> {
       await initWebpage(page, context, { url, autoconsent: true })
+
+      console.log('>  Extracting favicon...')
+      const faviconUrl = await getFaviconUrl(page)
+      const faviconDataUri = await faviconUrlToDataUri(faviconUrl, url)
```
zmarinov-astea: Can we merge the two favicon functions into one?

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +22,90 @@ const CONVERT_ITEM_PROPS = [
...
+
+  try {
+    const res = await fetch(url, { signal: AbortSignal.timeout(3000) })
+    if (!res.ok) return null
+
+    const buffer = await res.arrayBuffer()
+    const base64 = Buffer.from(buffer).toString('base64')
+    const contentType = res.headers.get('content-type') || 'image/x-icon'
+
+    return `data:${contentType};base64,${base64}`
```
zmarinov-astea, line 65: Is it a good idea to have the image data duplicated on each page @Sachanski ? Here they do it similarly though https://github.com/puppeteer/puppeteer/issues/2604

server/src/tasks/convert-to-pdf.ts
zmarinov-astea: https://www.mandatory.com/news/1809890-king-charles-prince-harry-feud-reconcile
image

server/src/tasks/convert-to-pdf.ts
strendafilov: Yes, we can, but these functions are separated because they have different responsibilities: getFaviconUrl() discovers the favicon URL in the browser context, while faviconUrlToDataUri() fetches the favicon and converts it to a data URI on the server. I separated them in case we decide to add more information from the page metadata later. If you prefer, I can merge them, but I don’t think it makes a big difference

zmarinov-astea: Ok, then we can write await faviconUrlToDataUri(await getFaviconUrl(page), url), or extract a separate function imageUrlToDataUri

server/src/tasks/convert-to-pdf.ts
```diff
@@ -81,16 +81,18 @@ function buildHeaderTemplate(faviconDataUri: string | null): string {
...
-    <div style="display: flex; align-items: center; justify-content: space-between; border-top: 1px solid #e2e4e8; padding-top: 6px;">
-      <span style="font-size: 10px; color: #b0b3ba; word-break: break-all; overflow-wrap: anywhere; max-width: 85%;" class="url"></span>
-      <span style="font-size: 8px; color: #9a9da3; flex-shrink: 0;">
-        <span class="pageNumber"></span> / <span class="totalPages"></span>
-      </span>
+function buildFooterTemplate(pageUrl: string): string {
+  return `
+    <div style="width: 100%; box-sizing: border-box; padding: 0 30px;">
+      <div style="display: flex; align-items: center; justify-content: space-between; border-top: 1px solid #e2e4e8; padding-top: 6px;">
+        <a href="${pageUrl}" style="font-size: 10px; color: #b0b3ba; text-decoration: underline; max-width: 85%; overflow: hidden; white-space: nowrap; text-overflow: ellipsis;">${pageUrl}</a>
```
zmarinov-astea, line 88: We can also make the link open in a new tab

zmarinov-astea (dismissed):

server/src/tasks/convert-to-pdf.ts
zmarinov-astea: We'll leave this for another task

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +27,90 @@ const CONVERT_ITEM_PROPS = [
...
+): Promise<string | null> {
+  const url = faviconUrl ?? `${new URL(pageUrl).origin}/favicon.ico`
+
+  try {
+    const res = await fetch(url, { signal: AbortSignal.timeout(3000) })
+    if (!res.ok) return null
+
+    const buffer = await res.arrayBuffer()
+    const base64 = Buffer.from(buffer).toString('base64')
+    const contentType = res.headers.get('content-type') || 'image/x-icon'
```
Sachanski: `.get` should be returning null if the header is not found so it will be better to use `??` instead of `||`

server/src/tasks/convert-to-pdf.ts
Sachanski: I think I have seen similar behaviour. I think it is OK as a first iteration. We can easiliy change it later on.

server/src/tasks/convert-to-pdf.ts
```diff
@@ -22,14 +27,90 @@ const CONVERT_ITEM_PROPS = [
...
+  `
+}
+
+function buildFooterTemplate(pageUrl: string): string {
+  return `
+    <div style="width: 100%; box-sizing: border-box; padding: 0 30px;">
+      <div style="display: flex; align-items: center; justify-content: space-between; border-top: 1px solid #e2e4e8; padding-top: 6px;">
+        <a href="${pageUrl}" style="font-size: 10px; color: #b0b3ba; text-decoration: underline; max-width: 85%; overflow: hidden; white-space: nowrap; text-overflow: ellipsis;">${pageUrl}</a>
+        <span style="font-size: 8px; color: #9a9da3; flex-shrink: 0;">
+          <span class="pageNumber"></span> / <span class="totalPages"></span>
```
Sachanski, line 90: I am assuming puppetter uses these empty spans and fills in th content?

Sachanski (dismissed):

server/src/tasks/convert-to-pdf.ts
strendafilov: Yes, puppeteer can provide us with the url, current date, title, current page and total pages.

Sachanski (approved):

PR #102, opened 2026-08-25
Fix input visualization in timeInput and page selector

strendafilov:


client/src/components/time-input/styles.module.css
```diff
@@ -16,7 +16,11 @@
     color: var(--theme-text-disabled);
   }
 
-  > form > input {
```
zmarinov-astea, line 19: I think the correct styling here should be 
```
.time-input {
    border: none;
    padding: 0;

    &[disabled] {
      color: var(--theme-text-disabled);
    }

    > input {
      width: 20px;
    }
  }
```

core-client/src/components/tapestry/items/pdf/page-selector/styles.module.css
```diff
@@ -4,9 +4,11 @@
   align-items: center;
 
   .page-input {
```
zmarinov-astea, line 6: Also here should be 
```
  .page-input {
    border-radius: 8px;
    padding: 5px 7px;
    > input {
      width: 25px;
    }
  }
```

zmarinov-astea:
We can try having a general **min-width: 0** for the input in core-client/src/components/lib/input, to avoid setting width on it every time, but I'm not sure if that will break other things...

client/src/components/time-input/styles.module.css
strendafilov: It makes a big difference actually, because in this way we remove the small boxes border, also the &[disabled] here won't work -`disabled` is on the nested <input>, not on .time-input itself. I think the expected result is to have a border around the time fields.

zmarinov-astea: No, this is how it was before, you can checkout an older version of tapestries. There wasn't a border around the time fields. I'm not sure about the disabled attribute, but in core-client/src/components/lib/input/styles.module.css, seems like disabled is being applied to the root which could also be incorrect.

zmarinov-astea (approved):

PR #103, opened 2026-08-25
Fix layering bugs

zmarinov-astea:


client/src/pages/tapestry/view-model/store-commands/items.ts
```diff
@@ -368,7 +368,7 @@ export function reorderItems(
       to === 'front' ? itemIds.has(i.dto.id) : !itemIds.has(i.dto.id),
     )
 
-    const allItems = [...sortBy(backItems, ['layer']), ...sortBy(frontItems, ['layer'])]
```
strendafilov, line 371: I think that it would be better to use a typed byLayer accessor instead of a string path 'dto.layer'
in order to ensure type safety. 
```
const byLayer = (item: typeof backItems[number]) => item.dto.layer
const allItems = [...sortBy(backItems, byLayer), ...sortBy(frontItems, byLayer)]
```

zmarinov-astea: Hm, we can create a typed version by using the Path<> type similar to useTapestryData, but I'm not sure how will that impact compilation performance

strendafilov: As far as I can see, Path<> makes sense when the field is dynamic passed in and you need generic type-safety over it. Here the field is statically known, so a simple typed accessor gives us the same safety without the compile-time cost that recursive Path<> can add.

zmarinov-astea: Yes, but we are using sortBy at different places, also we have an orderBy at one place

strendafilov: Okay, fair point.

Sachanski (approved):

PR #105, opened 2026-08-27
Fix refresh token logic

zmarinov-astea:
1. For some reason the request from the autorefresh timeout fails from time to time. 
2. If we are on the dashboard after the autorefresh failed, we continue to send the invalid(expired) access token, but don't receive 401, since the tapestries API and others also can accept requests from both authorized and unauthorized users. If we don't receive 401 a new refresh attempt is not triggered.
3. That is why I added an additional expiration check before using the access token on the client

### DIFF
--- a/client/src/pages/tapestry/view-model/socket-manager.ts
+++ b/client/src/pages/tapestry/view-model/socket-manager.ts
@@ -20,7 +20,7 @@ export type TapestryUpdated = TypedEvent<'tapestry-updated', TapestryUpdate>
 export class SocketManager extends TypedEventTarget<TapestryUpdated | RTCSignallerEvent> {
   private socket: Socket<ServerToClientEvents, ClientToServerEvents> = io(
     new URL(config.apiUrl).origin,
-    { path: SOCKET_PATH, autoConnect: false, auth: (cb) => cb({ token: auth.token }) },
+    { path: SOCKET_PATH, autoConnect: false, auth: (cb) => cb({ token: auth.accessToken?.token }) },
   )
   private isSignallerActivated = false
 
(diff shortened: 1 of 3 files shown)


Sachanski:
Any idea why the autorefresh fails?

Sachanski (approved):

zmarinov-astea:
> Any idea why the autorefresh fails?

I think it is most likely related to computer sleep/suspension 

>I have a set timeout which sends a network request. From time to time I see that this request did not complete successfully what could be the cause?
> When a network request inside a setTimeout fails from time to time, the root cause is rarely the network call itself. Instead, it is usually a timing mismatch caused by the way the browser handles memory, tabs, and operating system sleep cycles.Here are the most common reasons why a setTimeout network request fails to complete successfully:
> 1. The Computer Slept or Hibernated Mid-TimerIf a user closes their laptop lid or the computer goes to sleep while the setTimeout is waiting:The Problem: The timer stops ticking. When the computer wakes up, the timer resumes and immediately fires the network request.The Failure: Because the computer just woke up, the Wi-Fi card takes 2 to 5 seconds to re-establish a handshake with the router. The network request fires into a dead connection and instantly fails with a Network Error or Failed to fetch.
> 2. Tab Backgrounding / ThrottlingAs we discussed earlier, modern browsers aggressively throttle background tabs to save battery.The Problem: If the tab is minimized, a 10-second setTimeout might take 1 minute or longer to actually fire.The 
> 3. Page Navigation or Refresh (Aborted Requests)If the user clicks a link, submits a form, closes the tab, or refreshes the page while the network request is actively in flight
> 4. Overlapping Requests (Race Conditions)If the setTimeout is created recursively (e.g., a function that calls setTimeout on itself to poll data), a delayed background tab can queue them up. The browser or your backend server might rate-limit these simultaneous requests, leading to 429 Too Many Requests or dropped sockets.

PR #106, opened 2026-08-28
Enhancement/add link textbox

strendafilov:


client/src/components/link-action-modal/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onDelete: () => unknown
+}
+
+export function LinkActionModal({ link, onClose, onEdit, onDelete }: LinkActionModalProps) {
+  return (
+    <SimpleModal
+      title="Link"
+      onClose={onClose}
+      cancel={{ text: 'Edit', onClick: () => onEdit() }}
+      extraTrailingButtons={
```
zmarinov-astea: Can we use the confirm button instead of a new trailing one? Or swap the remove and edit buttons

client/src/components/message-input/index.tsx
```diff
@@ -33,6 +35,12 @@ export interface MessageInputProps {
 
 const TRAILING_EMPTY_PARAGRAPHS_REGEX = /(<p>(\s|<br\s*\/?>)*<\/p>)+$/gi
 
+interface LinkModalState {
```
zmarinov-astea: I saw this interface twice, and some duplicated logic at places. Is there a way to remove the duplication?

client/src/components/link-action-modal/index.tsx
strendafilov: The issue here is that I want the remove button to have a different style than the confirmation button. In case I swap them the remove button will be with primary color.

zmarinov-astea: Yes, but we can pass it additional styling with the confirm={{ ... }} attribute

strendafilov: Oh, yes. Okay I will change it, but should I also remove the different positioning of the additional buttons, or leave it?

client/src/components/message-input/index.tsx
strendafilov: I think there will be a way, but I can't come up with a scenario where I can devise a good abstraction.

client/src/components/link-action-modal/index.tsx
zmarinov-astea: I think we don't need them since the extra buttons were enough

client/src/components/message-input/index.tsx
zmarinov-astea: The easiest way to do it is to extract a hook useTextboxLink, which receives the editorAPI, and isEditMode, internally handles the LinkModalState and returns the modal UI as a single JSX variable, and also returns the addLink function. See client/src/components/item-picker/use-item-picker.tsx, should have similarities with it...

client/src/components/link-action-modal/index.tsx
zmarinov-astea: I mean yes, we should remove the different positioning and bring back the single extraButtons prop

client/src/components/link-action-modal/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+export function LinkActionModal({ link, onClose, onEdit, onDelete }: LinkActionModalProps) {
+  return (
+    <SimpleModal
+      title="Link"
+      onClose={onClose}
+      cancel={{ text: 'Edit', variant: 'outline', onClick: () => onEdit() }}
+      confirm={{ text: 'Remove', variant: 'primary-negative', onClick: () => onDelete() }}
+    >
+      <a href={link} target="_blank" rel="noopener noreferrer" style={{ wordBreak: 'break-all' }}>
+        {link}
```
zmarinov-astea, line 19: Should we handle internal links here?

client/src/components/message-input/index.tsx
```diff
@@ -79,8 +78,14 @@ export function MessageInput({
...
-    onLinkClick: handleCreateLink,
-    canAddLink: selectionState?.isLink,
+    onLinkClick: () => {
+      if (canAddLink) {
+        closeLinkModal()
+      } else {
+        addLink()
+      }
+    },
+    canAddLink,
```
zmarinov-astea: canAddLink should be renamed to addingLink after these changes.

client/src/components/tapestry-elements/items/text/index.tsx
```diff
@@ -36,49 +35,40 @@ export const TextItem = memo(({ id }: TapestryItemProps) => {
...
-
-    if (!anchorElement) {
-      return
-    }
-    setAddLinkProps({
-      content: editorAPI.current?.selectionText(),
-      element: domNode ?? anchorElement,
-    })
-  }
+  }, [isEditable, dispatch, id, unsavedContent, closeLinkModal])
```
zmarinov-astea: closeLinkModal should not be in the useEffect since this function is being created on every render and will trigger a replay of the useEffect on every render. We could write this instead outside the use effect
if (ui && !isEdtable) or if (addingLink && !isEditable) {
   closeLinkModal()
}

zmarinov-astea: Or use a closeLinkModalRef

client/src/components/link-action-modal/index.tsx
strendafilov: I think it's better to navigate away from the current page in this case. Imagine you don't know where a link leads and you want to check it; if it's an internal link, the modal would close and take you there, causing you to lose track of where you started. I think it's more user-friendly to open it in a new tab so you don't get lost.

zmarinov-astea: :+1:

client/src/hooks/use-textbox-link.tsx
```diff
@@ -0,0 +1,113 @@
...
+        onClose={close}
+        onApply={applyLink}
+      />
+    ) : null
+
+  return {
+    addLink,
+    closeLinkModal: close,
+    addingLink: !!linkModal,
+    isOpen: !!linkModal,
```
zmarinov-astea: We should choose addingLink or isOpen here

zmarinov-astea (dismissed):

client/src/hooks/use-textbox-link.tsx
strendafilov: I will remove isOpen, because it is not needed

zmarinov-astea (dismissed):

client/src/components/message-input/index.tsx
```diff
@@ -151,11 +156,24 @@ export function MessageInput({
...
                 onCreateLink: () => {
-                  handleCreateLink()
+                  addLink()
                   return true
                 },
+                onClick: (e) => {
+                  const maybeAnchor = iterateParents(
+                    e.target as HTMLElement,
+                    (el) => el.tagName !== 'A',
+                  )
```
Sachanski: I think we can remove this `iterateParents` function altogether and use `.closest` in its place

client/src/hooks/use-textbox-link.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const existingLink = editor.getAttributes('link').href as string | undefined
+    const initialLink = existingLink?.startsWith('?')
+      ? `${window.location.origin}${tapestryPath}${existingLink}`
+      : existingLink
+
+    setLinkModal({
+      mode: existingLink ? 'action' : 'edit',
+      initialText:
+        editorAPI.current?.selectionText().trim() &&
+        editorAPI.current.selectionText().trim() !== initialLink
```
Sachanski: I think we can either extract `editorAPI.current?.selectionText().trim()` into a constant or write the condition as `initialLink && editorAPI.current?.selectionText().trim() !== initialLink`.

client/src/hooks/use-textbox-link.tsx
```diff
@@ -0,0 +1,112 @@
...
+        onDelete={() => {
+          editorAPI.current?.editor().commands.unsetLink()
+          close()
+        }}
+      />
+    ) : linkModal?.mode === 'edit' ? (
+      <AddLinkModal
+        initialText={linkModal.initialText}
+        initialLink={linkModal.initialLink}
+        showTextField={true}
```
Sachanski: Just `showTextField`

client/src/hooks/use-textbox-link.tsx
```diff
@@ -39,13 +39,11 @@ export function useTextboxLink({
...
+    const selectionText = editorAPI.current?.selectionText().trim()
+
     setLinkModal({
       mode: existingLink ? 'action' : 'edit',
-      initialText:
-        editorAPI.current?.selectionText().trim() &&
-        editorAPI.current.selectionText().trim() !== initialLink
-          ? editorAPI.current.selectionText()
-          : undefined,
+      initialText: selectionText && selectionText !== initialLink ? selectionText : undefined,
```
Sachanski: This behaviour is slightly changed now - previously we were setting the `initialText` without trimming and only used trim while comparing. I don't know if that was intentional.

Also I meant that we can delete the entire `iterateParents` and replace all of its usages

strendafilov: Okay

Sachanski (approved):

PR #107, opened 2026-08-28
Delete item thumbnails after update.

zmarinov-astea:
Also fixed a thumbnail generation race condition, where an item could be edited while its thumbnail was being generated. After the first generation finished it reset  the scheduledThumbnailGeneration flag and the second generation could not regenerate the thumbnail.


server/src/tasks/generate-tapestry-thumbnails.ts
```diff
@@ -59,10 +62,14 @@ export async function generateTapestryThumbnails({
...
-        const { done, value } = await thumbnails!.next(item)
-        return done ? undefined : value
-      })
+      await processItemThumbnail(
+        item.id,
+        async () => {
+          const { done, value } = await thumbnails!.next(item)
+          return done ? undefined : value
+        },
+        (item.scheduledThumbnailProcessing || generateAll) === 'recreate',
```
Sachanski: `??` instead of `||`, also I am not sure how I feel about the parameter being called `generateAll` - sounds a bit like a boolean...

server/src/tasks/process-item-thumbnail.ts
```diff
@@ -81,75 +81,81 @@ function hasRenditionAtLOD(renditions: ImageAssetRendition[], lod: number) {
...
+          }),
+        )
+      }
+    }
+    completed = true
+  } finally {
+    if (!completed) {
+      await prisma.item.update({
+        where: { id: itemId },
+        data: { scheduledThumbnailProcessing: forceRegenerate ? 'recreate' : 'derive' },
```
Sachanski, line 157: Can't we do this in the catch block instead?

zmarinov-astea: No, because we have a return statement in the try block. Errors from the second processItemThumbnail call are handled differently in generateTapestryThumbnails, they stop the entire process of screenshoting,

server/src/tasks/generate-tapestry-thumbnails.ts
zmarinov-astea: ok what about generateAllProcess, genereateAllProcessing, generateAllMode?

Sachanski: `generationStrategy` or `generationMode` both sound find - it is mainly the verb that confuses me a little

Sachanski (dismissed):

Sachanski (approved):

PR #108, opened 2026-09-01
Support HEIC/HEIF image import

dbvisel:
## Summary

Dropping a `.heic`/`.heif` image currently fails with a generic "Could not import" error — no browser except Safari can decode HEIC, and even there the item would stay HEIC-sourced with no fallback for every other browser.

- Converts HEIC/HEIF images to JPEG server-side after upload, via a new worker job (`convert-heic-image.ts`) using libheif's `heif-convert` (added to the worker image via `apk add libheif-tools`).
- The job also corrects the item's `width`/`height` to the real decoded aspect ratio (using the same clamp/default-width math the client uses), since the client can't know the real size up front for a format it can't decode.
- The client shows a loading placeholder (reusing the existing `ItemPlaceholder`/`LoadingSpinner` pattern already used by the PDF viewer) during both the upload phase and the brief post-upload conversion window, so there's no broken-image flash.
- Scoped intentionally: only locally-uploaded (internally-hosted) sources are converted — an externally-hosted HEIC source is left alone, since converting it would silently turn a by-reference import into a copy. Only single-image HEIC/HEIF is handled — a multi-image sequence (burst/Live Photo) fails cleanly rather than partially.

## Verification

- `heif-convert` tested against a real iPhone HEIC photo, both in isolation and inside the actual built worker Docker image.
- Full end-to-end test against the real running app (drag-and-drop → placeholder → converted image, correct aspect ratio).
- `tsc --build`/eslint/prettier clean across `core`, `core-client`, `client`, `server`.

## Test plan

- [ ] Drag a `.heic`/`.heif` file onto a tapestry canvas
- [ ] Confirm a loading placeholder shows immediately (no broken-image icon)
- [ ] Confirm the image appears with the correct aspect ratio a few seconds later
- [ ] Confirm a normal (non-HEIC) image import is unaffected

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -35,6 +35,7 @@ RUN apk add --no-cache \
   yarn \
   ffmpeg \
   imagemagick \
+  libheif-tools \
   ghostscript;
 
 COPY --from=server /app /app
--- a/client/src/components/tapestry-elements/items/image/index.tsx
+++ b/client/src/components/tapestry-elements/items/image/index.tsx
@@ -1,5 +1,10 @@
 import { memo } from 'react'
-import { ImageItemViewer } from 'tapestry-core-client/src/components/tapestry/items/image/viewer'
+import {
+  HeicConvertingPlaceholder,
+  ImageItemViewer,
+} from 'tapestry-core-client/src/components/tapestry/items/image/viewer'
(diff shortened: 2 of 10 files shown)


Sachanski:
@dbvisel since we decided to do the conversion on the client I guess this one can be closed?

dbvisel:
Yes, we can close this!

PR #109, opened 2026-09-01
Support HEIC/HEIF image import (client-side conversion)

dbvisel:
## Summary

Dropping a `.heic`/`.heif` image currently fails with a generic "Could not import" error — no browser except Safari can decode HEIC.

This is an alternative to #108, converting entirely **client-side** instead of via a server-side worker job, for comparison:

- Uses [`heic-to`](https://github.com/hoppergee/heic-to) (LGPL-3.0, properly attributed, actively tracks upstream libheif releases), lazy-loaded via dynamic `import()` only when a HEIC file is actually dropped — confirmed via build output that it lands in its own chunk, never touching the main bundle.
- Conversion happens in a dedicated `ItemFactory` (`heicImageFactory`) *before* the tapestry item is created, converting the HEIC `File` to a JPEG `File` in the browser and only then handing it to the normal image item creation path. The item is never created with a placeholder size/UI — it appears once, with its real aspect ratio already known.
- No manual loading state: the existing `insertDataTransfer` wrapper already tracks `pendingRequests` around the whole import pipeline (for both file drops and URL pastes), so the standard hourglass indicator covers the conversion window automatically.
- **No server or Docker changes at all.**

## Tradeoffs vs. a server-side approach (#108)

- No dependency on worker availability — a busy worker (e.g. mid-screenshot job) can't delay this, since it never touches the job queue.
- Adds a real client bundle dependency (~700KB gzip, lazy-loaded) instead of a Docker image change.
- **Known limitation**: this only handles HEIC files dropped/selected directly as `File` objects. A pasted external `.heic` URL is not intercepted or converted — same as today's behavior for any URL-based import.

## Verification

- Tested end-to-end against the real running app (drag-and-drop → hourglass → converted image appears with correct aspect ratio, no empty/placeholder frame).
- `tsc --build`/eslint/prettier clean across `core`, `core-client`, `client`.

## Test plan

- [ ] Drag a `.heic`/`.heif` file onto a tapestry canvas
- [ ] Confirm the hourglass indicator shows while conversion is in progress, with no placeholder item visible
- [ ] Confirm the image appears with the correct aspect ratio once conversion finishes
- [ ] Confirm a normal (non-HEIC) image import is unaffected


client/src/stage/item-factories.ts
```diff
@@ -158,6 +159,13 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
   }
 }
 
+const heicImageFactory: ItemFactory = async (source, _mediaType, tapestryId) => {
+  if (!(source instanceof File) || !isHeicSource(source.name)) return null
```
zmarinov-astea: This is not a reliable way to check if the source is a heic image. We should first check the mediaType if it is image/heic or image/heif or other formats supported by heic-to, if not, proceed only if the file extension is heic or heif. We don't want to start the heic conversion for every file we drop. If the source is a link, download the image first, and then use maybe mediaSourceToBlob

core/src/utils.ts
```diff
@@ -160,3 +160,8 @@ export function determineImageFormat(url: string) {
 
   return mimeType.split('/')[1]
 }
+
+export function isHeicSource(source: string) {
+  const [, extension] = fileExtension(source.split(/[?#]/)[0])
```
zmarinov-astea: why are we splitting by ? or # when we are passing a file name?

client/src/stage/item-factories.ts
dbvisel: Fixed in b8b6246: now checks the already-resolved mediaType first (`image/heic`/`image/heif` — from File.type, a mime lookup on the filename, or the content-type proxy for a URL source), falling back to the extension only when mediaType can't be resolved. Also handles link/string sources now, not just dropped Files — downloads via `mediaSourceToBlob` before conversion.

core/src/utils.ts
dbvisel: Good catch — that stripping was leftover from an earlier design where this got called with a presigned URL. Simplified `isHeicSource` to only take a bare filename (b8b6246); the new HEIC factory now always reduces a URL source down to just its filename before calling it.

client/src/stage/item-factories.ts
```diff
@@ -159,10 +159,25 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
...
+
+function sourceFileName(source: MediaItemSource) {
+  if (source instanceof File) return source.name
+  return new URL(source).pathname.split('/').filter(Boolean).pop() ?? source
+}
+
+const heicImageFactory: ItemFactory = async (source, mediaType, tapestryId) => {
+  const isHeic = mediaType
+    ? HEIC_MEDIA_TYPES.includes(mediaType)
+    : isHeicSource(sourceFileName(source))
```
zmarinov-astea: This is again too complicated, if the given mediaType is not heic, check the filename extension only if the source is a File. Also we don't need the isHeicSource function now, we can just extract a constant array ['heic', 'hief'] for the extensions, put it below the heic media types array, and check it here.

client/src/stage/item-factories.ts
```diff
@@ -158,6 +159,28 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
...
+  if (!isHeic) return null
+
+  const file =
+    source instanceof File
+      ? source
+      : new File([await mediaSourceToBlob(source)], sourceFileName(source))
```
zmarinov-astea: Why are we creating a new file when the heic-to module can accept a simple Blob? Also when then returned file from convertHeicFile can have a random filename, it is not important, for example converted.jpg

client/src/stage/item-factories.ts
dbvisel: Fixed in ef1792e: the extension fallback now only applies to File sources — a URL falls back to nothing further since its mediaType already comes from the same content-type proxy either way. Also dropped isHeicSource entirely in favor of a plain HEIC_EXTENSIONS array checked against source.name, per your suggestion.

client/src/stage/item-factories.ts
dbvisel: Good catch, fixed in ef1792e — convertHeicFile now takes a Blob directly and always names the output converted.jpg, no more wrapping in a File just to carry a filename nothing used.

client/src/stage/item-factories.ts
```diff
@@ -158,6 +159,21 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
   }
 }
 
+const HEIC_MEDIA_TYPES = ['image/heic', 'image/heif']
+const HEIC_EXTENSIONS = ['heic', 'heif']
+
+const heicImageFactory: ItemFactory = async (source, mediaType, tapestryId) => {
+  const isHeic = mediaType
+    ? HEIC_MEDIA_TYPES.includes(mediaType)
```
zmarinov-astea: Can we use only HEIC_MEDIA_TYPES.includes for our isHeic check? The extension is already read at line 362 (mime.getType(source.name) ?? '') in client/src/model/data/utils.ts. See if that works properly on windows.

zmarinov-astea (dismissed):

client/src/stage/item-factories.ts
dbvisel: Fixed in 48cc3e4 — now relies on mediaType alone, no extension check in the factory at all.

On the Windows question: I don't have a Windows machine or a browser-automation tool to test this live, so I want to be upfront that this is research-backed, not empirically verified on a real device. What I found: without the Windows HEIF codec pack installed, Chromium reports an empty File.type for .heic (the common case — see the Chrome/HEIC issue reports on this), which is exactly the falsy case getMediaType() already falls through to mime.getType(source.name) for. I confirmed mime.getType() resolves .heic/.HEIC/mixed-case correctly regardless of platform. The one edge case I can't rule out: if some third-party software has written an incorrect MIME registration into a specific machine's Windows registry (a known general Windows pitfall, not HEIC-specific), File.type could report something misleading — but that's true independent of this change and isn't something client-side detection can defend against. Happy to dig further or have someone confirm on an actual Windows box if that's not reassuring enough.

zmarinov-astea (dismissed):

client/src/stage/item-factories.ts
```diff
@@ -160,9 +160,16 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
 }
 
 const HEIC_MEDIA_TYPES = ['image/heic', 'image/heif']
+const HEIC_EXTENSIONS = ['heic', 'heif']
+const GENERIC_MEDIA_TYPE = 'application/octet-stream'
 
 const heicImageFactory: ItemFactory = async (source, mediaType, tapestryId) => {
-  if (!mediaType || !HEIC_MEDIA_TYPES.includes(mediaType)) return null
+  const isHeic =
```
zmarinov-astea: Ok, a final cleanup, we can write `const isHeic = HEIC_MEDIA_TYPES.includes(mediaType && mediaType !== GENERIC_MEDIA_TYPE ? mediaType : source instanceof File ? mime.getType(source.name) : '')`

dbvisel: Done in 81b9e95 — using mime.getType(source.name) directly instead of a separate HEIC_EXTENSIONS array, same as getMediaType() already does for File sources.

zmarinov-astea (dismissed):

client/src/lib/heic.ts
```diff
@@ -0,0 +1,8 @@
+const HEIC_CONVERT_QUALITY = 0.92
+
+export async function convertHeicFile(blob: Blob): Promise<File> {
```
Sachanski: No need for the explicit return type here.

Also I think we can place this function in `lib/meida.ts` - we already have a `compressImage` function so it makes sense to put this one there as well

client/src/stage/item-factories.ts
```diff
@@ -158,6 +160,24 @@ const iaCollectionFactory: ItemFactory = async (source, _, tapestryId) => {
...
+const GENERIC_MEDIA_TYPE = 'application/octet-stream'
+
+const heicImageFactory: ItemFactory = async (source, mediaType, tapestryId) => {
+  const isHeic = HEIC_MEDIA_TYPES.includes(
+    mediaType && mediaType !== GENERIC_MEDIA_TYPE
+      ? mediaType
+      : source instanceof File
+        ? (mime.getType(source.name) ?? '')
+        : '',
+  )
```
Sachanski: Do we need this condition to be this complex? This looks a bit more concise and also we can remove the octet-stream mime type constant
```typescript
const isHeic = HEIC_MEDIA_TYPES.includes(
  source instanceof File ? (mime.getType(source.name) ?? '') : (mediaType ?? ''),
)
```

zmarinov-astea: The idea here was to handle the case where the extension is missing like on some Linux files or is incorrect

Sachanski: What about `(await fileTypeFromBlob(source))?.mime : ...` then?

client/src/lib/heic.ts
dbvisel: Done in 8f2afd3 - moved to media.ts next to compressImage, dropped the explicit return type.

client/src/stage/item-factories.ts
dbvisel: Went with this - fileTypeFromBlob reads the real file signature, so it works with no filename at all and needs no separate fallback for a missing or wrong extension. Verified against a real HEIC file with an empty name: still resolves to image/heic. Also drops the octet-stream special case entirely. Pushed in 8f2afd3.

Sachanski (approved):

Sachanski (approved):

PR #110, opened 2026-09-02
Rendering rels in groups 

### LINKED ISSUE #104: Can't select rels under groups
If a group has a background color or is selected, it is not possible to select or move rels under it. This should be possible as it is when rels are under a regular multiselection

strendafilov:


core-client/src/stage/renderer/index.ts
zmarinov-astea, line 227: We can return that a rel is selected if one of its anchor items is selected in order to make it moveable when it is under a selected group, also this way it's color won't be changed

core-client/src/stage/renderer/index.ts
```diff
@@ -224,14 +224,42 @@ export abstract class TapestryRenderer<
...
-        viewModel.dto.groupId &&
-        selection?.groupIds.has(viewModel.dto.groupId))
-    )
+      interactiveElement?.modelId === viewModel.dto.id
+    ) {
+      return true
+    }
+
+    if (
+      isItemViewModel(viewModel) &&
```
zmarinov-astea: Can we use the existing functions we have? isItemInSelection from core-client/src/view-model/utils.ts will do the same job

core-client/src/stage/renderer/index.ts
```diff
@@ -224,14 +224,42 @@ export abstract class TapestryRenderer<
...
+      viewModel.dto.groupId &&
+      selection?.groupIds.has(viewModel.dto.groupId)
+    ) {
+      return true
+    }
+
+    if (isRelViewModel(viewModel)) {
```
zmarinov-astea, line 247: Here we can instead extract and use a function isRelInSelection to core-client/src/view-model/utils.ts, and pass it the two adjasent items like this isRelInSelection(this.store.get(`items.${rel.dto.from.itemId}`), this.store.get(`items.${rel.dto.from.itemId}`), selection)

zmarinov-astea (approved):

Sachanski (approved):

PR #112, opened 2026-09-03
Add Openverse and Wikimedia Commons import

dbvisel:
## Summary

- Lets users paste a single-file page URL (image/PDF/video/audio) from Openverse or Wikimedia Commons to import that file directly, or a collection URL (an Openverse tag/source search, a Wikimedia Commons category) to open the existing bulk-import picker dialog.
- Both platforms share the same generalized proxy dispatch (`external-media` / `external-collection-count` / `external-collection-results`, each platform-tagged), the same `IAImport` union member (`ExternalCollection`), and the same picker list/details UI — while each keeps its own `core/src/<platform>.ts` module for the genuinely different API shapes (Openverse's numeric pagination vs. Commons' cursor-based one).
- Everything routes through the server-side proxy with caching, since both platforms rate-limit bursty client-side requests (confirmed live against both APIs).
- Commons video/audio resolve to WebM/MP3 transcodes instead of the original Ogg files, since modern browsers can't decode Ogg Theora and Safari can't decode Ogg Vorbis.
- The imported item's `notes` field records the originally pasted URL, since the resolved direct-file URL usually points somewhere else entirely (e.g. a third-party host for Openverse).

## Test plan

- [x] Paste a single Openverse image/audio page URL — creates the right item type, `notes` records the pasted URL
- [x] Paste an Openverse tag/source collection URL — picker opens, shows real count, scrolls/paginates, select-all respects the selection cap
- [x] Paste a single Wikimedia Commons `File:` page for each of image/PDF/video/audio — creates the right item type; video/audio resolve to browser-compatible transcodes
- [x] Paste a Wikimedia Commons category URL with a genuine mix of media types in one category — picker correctly dispatches per-item type, both individually and via bulk "Select all"
- [x] Confirmed live: real API rate-limiting on both platforms, cursor pagination across the real-page boundary and past the end of a category, and that a transient fetch failure no longer clears the picker or blocks retries for the full cache duration
- [x] `npm run lint` clean across `core`, `shared`, `server`, `client`
- [x] Manually tested end-to-end against a local Docker stack

https://claude.ai/code/session_017yQbFGogWMhhTHtVcz7PLn


server/src/resources/proxy.ts
zmarinov-astea, line 1: I don't think we need to route anything to the server, just to try avoiding the request limits. These imports don't happen that often and shouldn't trigger the restrictions. We could just disable the autoreload on our lazy list for the collection import

client/src/pages/tapestry/view-model/index.ts
zmarinov-astea, line 155: Since this type is no longer strictly connected to IA, can we rename it to CollectionImport? Also the external collection type should not exist, instead OpenverseCollection and wikimediaCommonsCategory should be direct subtypes

client/src/components/handle-ia-import-dialog/index.tsx
zmarinov-astea, line 1: handle-ia-import-dialog should be renamed to collection-import-dialog

client/src/components/handle-ia-import-dialog/import-items-list/external-collection-list/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+      checked={allSelected}
+      onChange={() => onToggleAll(!allSelected)}
+      total={total}
+      loading={toggling}
+      classes={{ root: mdOrLess ? styles.mobileSelectAll : undefined, checkbox: styles.checkbox }}
+      textVariant={textVariant}
+    />
+  )
+
+  return (
```
zmarinov-astea, line 174: Can we somehow extract a common collection-list component, which is used by all imports except playlist, which handles the ui and the lazy list. We can have additional components like wikimedia import which render the base component and handle more specific logic if necessary. Then we could pass items with unified optional props to the common one, or/and some additional grid column styling.

client/src/stage/item-factories.ts
```diff
@@ -170,6 +187,100 @@ const iaFactory: ItemFactory = async (source, _mediaType, tapestryId) => {
...
+  if (wikimediaTitle) {
+    const media = await fetchWikimediaMedia(wikimediaTitle)
+    if (!media) return null
+
+    const items = await createExternalMediaItems(tapestryId, [
+      { url: media.url, pageUrl: source, mediaType: media.mediaType },
+    ])
+    if (items.length === 0) return null
+
+    return { items, iaImports: [] }
```
zmarinov-astea, line 260: again, for this object type, iaImports should be renamed to collection imports

client/src/components/handle-ia-import-dialog/index.tsx
```diff
@@ -93,6 +162,26 @@ export function HandleIAImportDialog() {
...
+    })
+    try {
+      const viewModels = (await createNewItems(iaImport, selectedItems, tapestryId)).map(
+        createItemViewModel,
+      )
+      dispatch(viewModels.length > 0 && addAndPositionItems(viewModels))
+      onClose()
+    } finally {
+      dispatch((model) => {
+        model.pendingRequests--
```
zmarinov-astea, line 180: why are we modifying the pendingRequests here? We are not uploading anything.

client/src/components/handle-ia-import-dialog/import-details/index.tsx
```diff
@@ -72,3 +76,39 @@ export function ImportDetails({ import: iaImport }: ImportDetailsProps) {
...
+    collection.platform === 'openverse'
+      ? collection.collection.type === 'tag'
+        ? collection.collection.tag
+        : collection.collection.source
+      : collection.collection.category
+
+  const noun =
+    collection.platform === 'openverse'
+      ? collection.mediaType === 'image'
+        ? 'images'
```
zmarinov-astea, line 100: Create maps instead of having so many ternary operators here and on other places. What will happen when we import even more collection types...?

core/src/wikimedia-commons.ts
```diff
@@ -0,0 +1,277 @@
+const COMMONS_HOST = 'commons.wikimedia.org'
+const COMMONS_API_URL = 'https://commons.wikimedia.org/w/api.php'
+const COMMONS_PAGE_SIZE = 50
+const COMMONS_GENERIC_ICON_PATH = '/w/resources/assets/file-type-icons/'
+
```
zmarinov-astea, line 5: please leave links to the API documentation pages of commons and openverse

server/src/resources/proxy.ts
dbvisel: Removed the server-side proxy for this entirely. Both Openverse and Wikimedia Commons already send real CORS headers, so the client fetches them directly now. The actual fix for request volume is disabling the list's periodic background reload while the picker is open, not routing through our own server.

client/src/pages/tapestry/view-model/index.ts
dbvisel: Renamed to CollectionImport. Split ExternalCollection into direct OpenverseCollection and WikimediaCommonsCategory members, matching the existing IACollection/IAPlaylist pattern.

client/src/components/collection-import-dialog/index.tsx
dbvisel: Renamed to collection-import-dialog, including the exported component (HandleIAImportDialog to CollectionImportDialog).

client/src/components/collection-import-dialog/import-items-list/external-collection-list/index.tsx
```diff
@@ -0,0 +1,296 @@
...
+      checked={allSelected}
+      onChange={() => onToggleAll(!allSelected)}
+      total={total}
+      loading={toggling}
+      classes={{ root: mdOrLess ? styles.mobileSelectAll : undefined, checkbox: styles.checkbox }}
+      textVariant={textVariant}
+    />
+  )
+
+  return (
```
dbvisel: Added CollectionList. IASearchList and ExternalCollectionList now use it for the checkbox, the mobile/desktop row switch, and the LazyList wiring. Each caller only supplies its own thumbnail, detail columns, and CSS classes. Playlist entries stay separate, since they list a small fixed array rather than paging a remote collection.

client/src/stage/item-factories.ts
dbvisel: Renamed to collectionImports here too, for the same reason.

client/src/components/collection-import-dialog/index.tsx
```diff
@@ -93,6 +162,26 @@ export function HandleIAImportDialog() {
...
+    })
+    try {
+      const viewModels = (await createNewItems(iaImport, selectedItems, tapestryId)).map(
+        createItemViewModel,
+      )
+      dispatch(viewModels.length > 0 && addAndPositionItems(viewModels))
+      onClose()
+    } finally {
+      dispatch((model) => {
+        model.pendingRequests--
```
dbvisel: This dialog's confirm handler creates items directly. It does not go through the drop/paste pipeline. That pipeline wraps every import in a pendingRequests increment, so it gets the loading indicator for free. This handler needs the same wrapping by hand.

This matters in practice. Creating a few pdf items from a picker runs getPDFItemSize, which downloads and parses the whole file client-side. A real multi-megabyte PDF from a Commons category took about ten seconds with no visible feedback before this fix.

client/src/components/handle-ia-import-dialog/import-details/index.tsx
dbvisel: Replaced the ternaries with per-type components and config objects (OpenverseCollectionDetails, WikimediaCommonsCategoryDetails, and a describeExternalCollection function). Adding a platform now means adding one branch in one place, not touching every ternary that reads the platform field.

core/src/wikimedia-commons.ts
dbvisel: Added links to both APIs' documentation.

client/src/components/collection-import-dialog/import-items-list/ia-search-list/index.tsx
zmarinov-astea, line 1: this folder should be renamed to ia-search-list

client/src/components/collection-import-dialog/import-items-list/collection-list/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  classes,
+  shouldRenderItem,
+  renderItemContent,
+  renderItemDetails,
+  isSelected,
+  onSelectItem,
+  selectedCount,
+  header,
+  emptyPlaceholder,
+}: CollectionListProps<T>) {
```
zmarinov-astea, line 129: Can you use the LazyListProps interface to derive the prop types here? Also you can use a rest ...props parameter

client/src/components/collection-import-dialog/import-items-list/external-collection-list/index.tsx
zmarinov-astea, line 1: What I meant was to split this external collection list into wikimedia list and openverse list components

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -72,3 +80,66 @@ export function ImportDetails({ import: iaImport }: ImportDetailsProps) {
...
   )
 }
+
+type OpenverseCollectionImport = Extract<CollectionImport, { type: 'OpenverseCollection' }>
+type WikimediaCommonsCategoryImport = Extract<
+  CollectionImport,
+  { type: 'WikimediaCommonsCategory' }
+>
+
+function CollectionDetailsLayout({
```
zmarinov-astea: Openverse and wikimedia should not have a common collection details layout, since we can extract more details from the commons category like description and image, which could be different from openverse's. Conversely, if they are indeed to similar then we should have a GenericDetailsLayout component which could be also used by the IA search

client/src/components/collection-import-dialog/import-items-list/ia-playlist/index.tsx
zmarinov-astea, line 1: This folder should be ia-playlist now

client/src/components/collection-import-dialog/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+    setSelectedItems([])
+    if (importIndex === collectionImports.length - 1) {
+      dispatch(setCollectionImports([]))
+      setImportIndex(0)
+    } else {
+      setImportIndex(importIndex + 1)
+    }
+  }
+
+  const header = <ImportDetails import={collectionImport} />
```
zmarinov-astea: header is not a good variable name for this

client/src/components/collection-import-dialog/import-items-list/external-collection-list/styles.module.css
```diff
@@ -0,0 +1,99 @@
...
+    align-items: center;
+    grid-template-columns: auto repeat(var(--detail-column-count), 130px);
+
+    @media screen and (width <= 900px) {
+      padding: 5px var(--padding-x);
+      grid-template-columns: auto 15px;
+    }
+
+    .item-image {
+      height: 30px;
```
zmarinov-astea: Can we set the thumbnail size to 60x60? Same for ia items

client/src/stage/item-factories.ts
```diff
@@ -145,31 +166,123 @@ const iaFactory: ItemFactory = async (source, _mediaType, tapestryId) => {
...
+      } catch {
+        return null
+      }
+    }),
+  )
+
+  return compact(items)
+}
+
+const externalMediaFactory: ItemFactory = async (source, _mediaType, tapestryId) => {
```
zmarinov-astea: This looks like it ca be split into openverse factory and wikimedia factory

core/src/wikimedia-commons.ts
```diff
@@ -0,0 +1,280 @@
...
+  try {
+    const url = new URL(COMMONS_API_URL)
+    url.searchParams.set('action', 'query')
+    url.searchParams.set('generator', 'categorymembers')
+    url.searchParams.set('gcmtitle', collection.category)
+    url.searchParams.set('gcmlimit', String(COMMONS_PAGE_SIZE))
+    url.searchParams.set('gcmtype', 'file')
+    url.searchParams.set('prop', 'videoinfo')
+    url.searchParams.set('viprop', VIDEO_INFO_PROPS)
+    url.searchParams.set('viurlwidth', '300')
```
zmarinov-astea: why are we looking for video collection items? Also can we request a smaller thumbnail close to 60x60 what is displayed in the UI?

client/src/components/collection-import-dialog/import-items-list/ia-search-list/index.tsx
dbvisel: Done. Renamed to ia-search-list.

client/src/components/collection-import-dialog/import-items-list/collection-list/index.tsx
dbvisel: Done. CollectionListProps now Picks the shared fields from LazyListProps, and forwards them to LazyList via a rest parameter instead of naming each one twice.

client/src/components/collection-import-dialog/import-items-list/external-collection-list/index.tsx
dbvisel: Done. Split into OpenverseCollectionList and WikimediaCollectionList, each with its own concrete item type. Both still render through the shared CollectionList.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Checked the real Commons API before deciding. Category pages have no usable description -- extracts (Wikipedia's article-summary endpoint) always returns an empty string for them, since Commons categories are templated, not prose. pageimages does resolve a real representative thumbnail, though. So: separated the two (no more shared layout), kept Openverse as-is (label + total, it has no natural image), and added the real category thumbnail to Wikimedia's.

client/src/components/collection-import-dialog/import-items-list/ia-playlist/index.tsx
dbvisel: Done. Renamed to ia-playlist.

client/src/components/collection-import-dialog/index.tsx
dbvisel: Renamed to importDetails.

client/src/components/collection-import-dialog/import-items-list/external-collection-list/styles.module.css
dbvisel: Done. 60x60 for both.

client/src/stage/item-factories.ts
dbvisel: Done. Split into openverseFactory and wikimediaFactory.

core/src/wikimedia-commons.ts
dbvisel: We're not filtering to video -- prop=videoinfo is MediaWiki's general file-metadata endpoint for Commons, despite the name. Verified with a real API call against a plain JPEG: it comes back under videoinfo with mediatype: BITMAP, full url/mime/thumburl included. imageinfo lacks the derivatives field this code needs for video/audio playback URLs, so videoinfo is the one prop that works for every file type here, not just video.

Also reduced viurlwidth from 300 to 120 (2x the 60px now displayed).

client/src/components/collection-import-dialog/import-items-list/openverse-collection-list/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+            </>
+          ) : (
+            header
+          )
+        }
+        shouldRenderItem={(item) => !undecodableIds.has(item.id)}
+        isSelected={(item) => !!selectedItems.find((i) => i.id === item.id)}
+        onSelectItem={(item) => onSelect({ id: item.id, sourceUrl: item.url })}
+        selectedCount={selectedCount}
+        renderItemContent={(item) => (
```
zmarinov-astea: Can you extract all item row rendering logic into collection-list. The UI should be unified so you can create a common interface for the item having optional properties - image, title, detials, downloads, creator, etc. and maybe a boolean map for specifying which columns to display. No need for import specific UI like NO_THUMBNAIL_ICON,

client/src/components/collection-import-dialog/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+export function CollectionImportDialog() {
+  const { collectionImports, id: tapestryId } = useTapestryData(['collectionImports', 'id'])
+  const dispatch = useDispatch()
+  const [selectedItems, setSelectedItems] = useState<ImportItem[]>([])
+  const mdOrLess = useResponsive() <= Breakpoint.MD
+
+  const [importIndex, setImportIndex] = useState(0)
+  const collectionImport = collectionImports[importIndex] as CollectionImport | undefined
+
+  const { trigger: toggleAll, loading } = useAsyncAction(async ({ signal }, check: boolean) => {
```
zmarinov-astea: Since we don't reload the list anymore, create a mechanism which gets the first MAX_SELECTION items from the list itself, so we don't fetch them on two places, or maybe it sets them here when its ready. Also the select all button could be disabled before the first result comes,

client/src/lib/wikimedia.ts
zmarinov-astea, line 1: This file is called external media but handles only wikimedia, rename it to something more accurate like wikimedia manager or wikimedia cursor service or just wikimedia

client/src/components/collection-import-dialog/import-items-list/request-external-items.ts
```diff
@@ -0,0 +1,60 @@
...
+      clearTimeout(timeout)
+      reject(new DOMException('Aborted', 'AbortError'))
+    })
+  })
+}
+
+// LazyListLoader replaces the whole list on every fetch. This includes its own periodic reload.
+// A single failed fetch would then clear the list. Retry first. Space the retries past the
+// proxy's short failure-cache time.
+async function fetchPageWithRetry<Result>(
```
zmarinov-astea: This retry logic complicates the process too much we don't need it

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+const parser = new DOMParser()
+
+interface ImportDetailsProps {
+  import: CollectionImport
+}
+
+// TODO: Extract a shared layout component. This removes the duplication between the
+// branches below.
+export function ImportDetails({ import: collectionImport }: ImportDetailsProps) {
```
zmarinov-astea: Can we now unify the layout of all these? We could again use an interface with optional props for the details contents.

client/src/components/collection-import-dialog/import-items-list/openverse-collection-list/index.tsx
dbvisel: Done. CollectionList now owns all item-row rendering (thumbnail/fallback icon, title, detail columns), plus the desktop column-header bar and the root/mobile-select-all scaffolding, which were already byte-for-byte identical across all three lists. Each caller maps its own native item type into a new CollectionListItem shape via a toListItem prop -- platform-specific bits like the thumbnail fallback-icon map stay local to each list (not exposed on the shared interface), matching the earlier ask to keep import-specific UI like NO_THUMBNAIL_ICON out of collection-list.

client/src/components/collection-import-dialog/index.tsx
dbvisel: Done both ways. Each list now triggers its own already-defined requestItems function directly for select-all and reports the mapped items up via a new onSelectAll callback, instead of the dialog re-implementing each platform's fetch separately -- one fetch implementation per platform, not two. Also disabled the select-all checkbox until total is known (SelectAll's Checkbox is now disabled when total === undefined).

client/src/lib/wikimedia.ts
dbvisel: Done. Renamed to wikimedia.ts.

client/src/components/collection-import-dialog/import-items-list/request-external-items.ts
dbvisel: Done. Removed the retry/delay logic entirely -- requestExternalItems just calls fetchPage once now.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. Extracted a shared DetailsLayout (thumbnail/title/subtitle/meta/body, all optional except title) used by every branch -- Openverse, Commons (still its own component, since it owns the async thumbnail fetch), IA search, and the generic IA item/collection case.

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      {body !== undefined && (
+        <Text variant={textVariant} component="div">
+          {body}
+        </Text>
+      )}
+    </div>
+  )
+}
+
+interface ImportDetailsProps {
```
zmarinov-astea: Move the props interface directly before the function component

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      subtitle={`${collection.total} files`}
+    />
+  )
+}
+
+export function ImportDetails({ import: collectionImport }: ImportDetailsProps) {
+  const mdOrLess = useResponsive() <= Breakpoint.MD
+  const textVariant = mdOrLess ? 'bodyXs' : undefined
+
+  if (collectionImport.type === 'OpenverseCollection') {
```
zmarinov-astea: use a switch statement here

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+  import: CollectionImport
+}
+
+type OpenverseCollectionImport = Extract<CollectionImport, { type: 'OpenverseCollection' }>
+type WikimediaCommonsCategoryImport = Extract<
+  CollectionImport,
+  { type: 'WikimediaCommonsCategory' }
+>
+
+const OPENVERSE_MEDIA_TYPE_NOUN: Record<OpenverseCollectionImport['mediaType'], string> = {
```
zmarinov-astea: Dont' use a map only when we have two values here

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+    </div>
+  )
+}
+
+interface ImportDetailsProps {
+  import: CollectionImport
+}
+
+type OpenverseCollectionImport = Extract<CollectionImport, { type: 'OpenverseCollection' }>
+type WikimediaCommonsCategoryImport = Extract<
```
zmarinov-astea: These two types are used only once so remove them and directly type the function parameter instead

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+  CollectionImport,
+  { type: 'WikimediaCommonsCategory' }
+>
+
+const OPENVERSE_MEDIA_TYPE_NOUN: Record<OpenverseCollectionImport['mediaType'], string> = {
+  image: 'images',
+  audio: 'audio items',
+}
+
+function openverseCollectionLabel(collection: OpenverseCollectionImport['collection']): string {
```
zmarinov-astea: This function is also not necessary

client/src/components/collection-import-dialog/import-details/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+function WikimediaCommonsCategoryDetails({
+  collection,
+}: {
+  collection: WikimediaCommonsCategoryImport
+}) {
+  const [thumbnail, setThumbnail] = useState<string | null>(null)
+
+  useEffect(() => {
+    const controller = new AbortController()
+    void fetchWikimediaCollectionThumbnail(collection.collection, controller.signal).then(
```
zmarinov-astea: We have a useAsync for this operation

client/src/components/collection-import-dialog/import-items-list/collection-list/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * The UI and lazy-loading shell shared by every collection import's item list (every
+ * `ImportItemsList` branch except `IAPlaylistEntries`, which lists a fixed, already-loaded
+ * array rather than paging a remote collection). Each caller supplies its own fetching and a
+ * `toListItem` mapper into the generic `CollectionListItem` shape; item-row rendering (thumbnail
+ * or fallback icon, title, detail columns), the checkbox wiring, the mobile details/desktop row
+ * switch, the desktop column-header bar, and the `LazyList` plumbing all live here once.
+ */
+export interface CollectionListProps<T extends WithId> extends Pick<
```
zmarinov-astea: remove this huge comment and rename the component to BaseCollectionList and also the folder name

client/src/lib/media.ts
```diff
@@ -23,6 +23,10 @@ export function mediaSourceToBlob(source: MediaItemSource) {
   return source instanceof File ? source : urlToBlob(source)
 }
 
+export function resolveExternalMediaSource(url: string): Promise<MediaItemSource> {
```
zmarinov-astea: This function is not necessary

client/src/stage/data-transfer-handler.ts
```diff
@@ -38,7 +38,7 @@ export async function parseMediaSource(
     }
   }
 
-  return { items: [], iaImports: [] }
+  return { items: [], collectionImports: [] }
```
zmarinov-astea: can we make this collectionImports prop optional?

core/src/wikimedia-commons.ts
```diff
@@ -0,0 +1,312 @@
...
+    return null
+  }
+}
+
+export interface WikimediaCursorStore {
+  get(realPage: number): Promise<string | null | undefined>
+  set(realPage: number, cursor: string | null): Promise<void>
+}
+
+export async function fetchWikimediaCollectionResults(
```
zmarinov-astea: Simplify this API by having only the optional cursor value as parameter, not the whole store, page and pageSize

core/src/wikimedia-commons.ts
```diff
@@ -0,0 +1,312 @@
...
+  mediaType: WikimediaMediaType
+}
+
+interface CommonsDerivative {
+  src: string
+  type: string
+  width: number
+}
+
+interface CommonsVideoInfo {
```
zmarinov-astea: This interface name is misleading, since prop=videoinfo just extends the regular imageinfo result to have video props. rename ite to CommonsItemInfo, also everywhere you have used video where it refers to the general wikimedia item rename it to something else

core/src/wikimedia-commons.ts
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// Build a short link back to a file's page from just its page id. This avoids re-encoding a
+// title with unicode, spaces, or punctuation. Commons' own API reports this exact form as
+// `descriptionshorturl` for every file.
+export function wikimediaFilePageURL(pageId: string): string {
+  return `https://${COMMONS_HOST}/w/index.php?curid=${pageId}`
+}
+
+export type WikimediaCollectionQuery = { type: 'category'; category: string }
```
zmarinov-astea: since we are quierying only categories for now remove this type

client/src/components/collection-import-dialog/import-items-list/wikimedia-collection-list/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const mdOrLess = useResponsive() <= Breakpoint.MD
+  const textVariant = mdOrLess ? 'bodyXs' : undefined
+
+  const [listLoader, setListLoader] = useState<LazyListLoader<WikimediaMedia> | null>(null)
+  const state = useObservable(listLoader)
+  const total = state?.total
+
+  const [undecodableIds, setUndecodableIds] = useState<Set<string>>(new Set())
+  const [loadFailed, setLoadFailed] = useState(false)
+
```
zmarinov-astea, line 59: in this file is where you should keep track of wikipedia's page sizes and the current cursor, instead of doing complicated math for page and pageSize

client/src/components/collection-import-dialog/import-items-list/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+  onSelect: (item: ImportItem) => unknown
+  onSelectAll: (items: ImportItem[]) => unknown
+  onDeselectAll: () => unknown
+  collectionImport: CollectionImport
+  selectedItems: ImportItem[]
+  header?: ReactNode
+}
+
+export function ImportItemsList({ collectionImport, ...props }: ImportItemsListProps) {
+  if (collectionImport.type === 'IACollection') {
```
zmarinov-astea: can we use a switch statement here

client/src/components/collection-import-dialog/import-items-list/request-external-items.ts
```diff
@@ -0,0 +1,26 @@
+import { paginateBySkipLimit } from './paginate-by-skip-limit'
+
+export async function requestExternalItems<Media>(
```
zmarinov-astea: can you move this and paginate-by-skip-limit to a paginate-utils file?

client/src/components/collection-import-dialog/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+              pageUrl: openverseMediaPageURL(mediaType, id),
+              mediaType,
+            },
+        ),
+      ),
+    )
+  }
+
+  if (collectionImport.type === 'WikimediaCommonsCategory') {
+    return createExternalMediaItems(
```
zmarinov-astea, line 55: Can you use only one createExternalmediaItems call and a ternary operator in the items map

client/src/pages/tapestry/view-model/index.ts
```diff
@@ -162,6 +164,17 @@ export type IAImport =
...
     }
+  | {
+      type: 'OpenverseCollection'
+      mediaType: OpenverseMediaType
+      collection: OpenverseCollectionQuery
+      total: number
+    }
+  | {
+      type: 'WikimediaCommonsCategory'
+      collection: WikimediaCollectionQuery
```
zmarinov-astea: Wikimedia does not have collection, so we can just have category title here

client/src/components/collection-import-dialog/import-items-list/wikimedia-collection-list/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      const result = await requestItems(0, MAX_SELECTION, signal)
+      onSelectAll(result.data.map(toImportItem))
+    },
+  )
+
+  const selectedCount = selectedItems.length
+  const maxSelectable = total === undefined ? undefined : Math.min(total, MAX_SELECTION)
+  const allSelected = maxSelectable !== undefined && selectedCount >= maxSelectable
+
+  const selectAll = (
```
zmarinov-astea, line 97: Is there a way to omit the repetition of selectAll in every list file?

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. Moved ImportDetailsProps directly above ImportDetails.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. ImportDetails now uses a switch on collectionImport.type.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. Replaced OPENVERSE_MEDIA_TYPE_NOUN with a ternary.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. Removed both types -- the switch's own narrowing covers Openverse's case, and WikimediaCommonsCategoryDetails now types its collection prop inline with Extract<CollectionImport, {type: 'WikimediaCommonsCategory'}>.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. Inlined openverseCollectionLabel at its one call site.

client/src/components/collection-import-dialog/import-details/index.tsx
dbvisel: Done. WikimediaCommonsCategoryDetails now uses useAsync instead of manual useEffect/useState/AbortController.

client/src/components/collection-import-dialog/import-items-list/collection-list/index.tsx
dbvisel: Done. Removed the comment, renamed the component to BaseCollectionList and the folder to base-collection-list.

client/src/lib/media.ts
dbvisel: Done. Removed resolveExternalMediaSource -- it was a no-op identity wrapper (url => Promise.resolve(url)). createExternalMediaItems now passes the URL straight to createMediaItem.

client/src/stage/data-transfer-handler.ts
dbvisel: Done. ItemFactoryResult.collectionImports is now optional, so every factory/branch that never produces one just returns {items} -- dropped the collectionImports: [] boilerplate throughout item-factories.ts and data-transfer-handler.ts.

core/src/wikimedia-commons.ts
dbvisel: Done. fetchWikimediaCollectionResults now takes (category, cursor?, signal?) and returns {results, nextCursor} -- no more page/pageSize/cursorStore.

core/src/wikimedia-commons.ts
dbvisel: Done. Renamed to CommonsItemInfo, and the videoInfo local variable / VIDEO_INFO_PROPS constant that referred to it throughout (now itemInfo / ITEM_INFO_PROPS).

core/src/wikimedia-commons.ts
dbvisel: Done. Removed WikimediaCollectionQuery -- every function now takes the category string directly, and CollectionImport's WikimediaCommonsCategory member has a plain category: string field instead of collection.

client/src/components/collection-import-dialog/import-items-list/wikimedia-collection-list/index.tsx
dbvisel: Done. wikimedia-collection-list now keeps a flat buffer of everything fetched so far plus the next cursor, fetching one more real page whenever the buffer doesn't yet cover the requested skip/limit window -- no page-size math at all.

client/src/components/collection-import-dialog/import-items-list/index.tsx
dbvisel: Done. ImportItemsList now uses a switch on collectionImport.type.

client/src/components/collection-import-dialog/import-items-list/request-external-items.ts
dbvisel: Done. Merged into paginate-utils.ts.

client/src/components/collection-import-dialog/index.tsx
dbvisel: Done. Both branches now go through one createExternalMediaItems call, deciding pageUrl/mediaType per item via whether wikimediaMediaType is present.

client/src/pages/tapestry/view-model/index.ts
dbvisel: Done. WikimediaCommonsCategory now has a plain category: string field.

client/src/components/collection-import-dialog/import-items-list/wikimedia-collection-list/index.tsx
dbvisel: Done. Added CollectionSelectAll (in base-collection-list) wrapping SelectAll with the classes/textVariant every caller needs -- each list now just passes checked/onChange/total/loading/mdOrLess.

PR #120, opened 2026-09-10
Bugfix - unable to select rels

strendafilov:


core-client/src/components/tapestry/tapestry-canvas/index.tsx
```diff
@@ -80,15 +80,10 @@ function TapestryElementLocator({
...
-  if (!item) {
-    return null
-  }
   const isInteractive = id === interactiveElement?.modelId
-  const isInSelection = isItemInSelection(item, selection)
-  const displayDom = shouldDisplayDom(
-    { disableOptimizations, thumbnailsInitialized, isInteractive },
-    item,
-  )
+  const isInSelection = !!item && isItemInSelection(item, selection)
```
zmarinov-astea: there is no problem to pass undefined here

core-client/src/components/tapestry/tapestry-canvas/index.tsx
```diff
@@ -80,15 +80,10 @@ function TapestryElementLocator({
...
+  const displayDom =
+    !item || shouldDisplayDom({ disableOptimizations, thumbnailsInitialized, isInteractive }, item)
```
zmarinov-astea, line 86: Maybe the logic should be a little more complex here: item ? shouldDisplayDom(...) : isInteractive, so we don't render empty divs, but we can leave it like that for now.

zmarinov-astea (dismissed):

core-client/src/components/tapestry/tapestry-canvas/index.tsx
strendafilov: Ok, then I'll remove it

zmarinov-astea (approved):

PR #122, opened 2026-09-11
Bugfix: deleting elements 

### LINKED ISSUE #125: The elements are restored after being deleted
Here is an example which explains the issue: We move an element from left to right, pause, wait for the mutation request to be sent and immediately delete it. When the response from the first request returns, the element reappears (this happens most often with a slow internet connection). This results in the resurrection of elements, even though they have already been deleted once.

strendafilov:
Resolved this issue: rapid deletions where a resource could be resurrected by an in-flight server response. During mutation response handling, incoming data is ignored if the resource was deleted after the push request started.

### DIFF
--- a/client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
+++ b/client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
@@ -141,6 +141,28 @@ function createItemPatch(newItem: ItemDto, oldItem: ItemDto) {
   return isEmpty(patch) ? undefined : { ...patch, type: newItem.type }
 }
 
+const ITEM_READONLY_PROPS = [
+  'createdAt',
+  'updatedAt',
+  'tapestry',
+  'scheduledThumbnailProcessing',
+  'thumbnail',
+] as const
+
+const MEDIA_ITEM_READONLY_PROPS = [...ITEM_READONLY_PROPS, 'internallyHosted'] as const
+
+function itemToCreateParams(item: ItemDto): ItemCreateDto & { id: string } {
+  const readonlyProps: readonly string[] = isMediaItem(item)
+    ? MEDIA_ITEM_READONLY_PROPS
+    : ITEM_READONLY_PROPS
+
+  const createParams = Object.fromEntries(
+    Object.entries(item).filter(([key]) => !readonlyProps.includes(key)),
+  ) as ItemCreateDto & { id: string }
+
+  return createParams
+}
+
 type EventTypesMap = {
   socketManager: EventTypes<SocketManager>
 }
@@ -246,7 +268,7 @@ export class TapestryResourcesRepo extends ResourceRepo<TapestryResourceName, Ty
       throw new Error('Cannot create new tapestries!')
     }
     if (resourceName === 'items') {
-      return resource as ItemCreateDto & { id: string }
+      return itemToCreateParams(resource as ItemDto)
     }
     if (resourceName === 'groups') {
       return resource as GroupCreateDto & { id: string }
--- a/client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
+++ b/client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
@@ -27,7 +27,7 @@ import {
   RelUpdateDto,
 } from 'tapestry-shared/src/data-transfer/resources/dtos/rel'
 import { idFilter, listAll, resource, RESTMethodOptions } from '../../../services/rest-resources'
-import { isEmpty, isEqual } from 'lodash-es'
+import { isEmpty, isEqual, omit } from 'lodash-es'
 import {
   BaseResourceDto,
   BatchMutationDto,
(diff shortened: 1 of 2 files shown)


client/src/utils/resource-repo.ts
```diff
@@ -326,10 +350,17 @@ export abstract class ResourceRepo<
...
       toSetInWorkingCopy.forEach((resource) => {
         const resourcePath = [resourceName, resource.id]
+        const deletedAt = get(this.deletedResourceVersions, resourcePath) as number | undefined
+        const wasDeletedAfterThisCommit = deletedAt !== undefined && deletedAt > commitVersion
+
         // Don't override local changes that may have happened since the push operation started
-        if (get(this.resourceVersions, resourcePath, 0) <= commitVersion) {
+        if (
+          !wasDeletedAfterThisCommit &&
+          get(this.resourceVersions, resourcePath, 0) <= commitVersion
```
zmarinov-astea, line 359: Is it necessary to keep a separate object for deleted resource versions? If the object is deleted, its version does not exist at all. The bug gets solved on my machine by only adding an additional check if the version exists. Also removed the 0 fallback: 
```
// Don't override local changes that may have happened since the push operation started
        if (
          get(this.resourceVersions, resourcePath) !== undefined &&
          get(this.resourceVersions, resourcePath) <= commitVersion
        ) {
```
See if that works on your side.

client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
```diff
@@ -246,7 +261,7 @@ export class TapestryResourcesRepo extends ResourceRepo<TapestryResourceName, Ty
       throw new Error('Cannot create new tapestries!')
     }
     if (resourceName === 'items') {
-      return resource as ItemCreateDto & { id: string }
+      return itemToCreateParams(resource as ItemDto)
```
zmarinov-astea, line 264: Why do we need to clear the readonly props? The diff function is not necessarily tied to sending the new data to the server.

client/src/utils/resource-repo.ts
zmarinov-astea, line 367: If the undo after deletion bug still persists, I think it should be handled here. If the deletion succeeds on the server but the item gets restored locally, we should check the resource version similar to toSeInWorkingCopy and apply the deletion only if it exists and is lower.

client/src/utils/resource-repo.ts
strendafilov: I have tested it and it works in all cases so I will remove the logic with deletedResourceVersions object.

client/src/pages/tapestry/view-model/tapestry-resources-repo.ts
strendafilov: createParams() isn't just used for local diffing - its return value is sent directly to the server as the create payload in mutate() (see itemBatchMutations). createdAt, thumbnail, scheduledThumbnailProcessing, internallyHosted are readonly fields, so if we don't strip them we risk the API rejecting the request or silently overwriting values ​​it's supposed to own. So I think that itemToCreateParams needs to stay.

zmarinov-astea: Ok, but the tapestryId and id are also readonly props, and we shouldn't exclude them. Also the server is designed to deliberately overwrite createdAt and updatedAt, look at server/src/resources/items.ts

client/src/utils/resource-repo.ts
```diff
@@ -326,16 +326,22 @@ export abstract class ResourceRepo<
...
+        ) {
           // eslint-disable-next-line @typescript-eslint/no-floating-promises
           set(workingCopy, resourcePath, resource)
         }
       })
 
       toRemoveFromWorkingCopy.forEach((id) => {
-        delete (workingCopy as ResourceIdMaps<R, TypeMap>)[resourceName][id]
-        delete this.resourceVersions[resourceName][id]
+        if (get(this.resourceVersions, [resourceName, id], 0) <= commitVersion) {
```
zmarinov-astea, line 341: Maybe we should put a comment here that a higher version after deletion is possible when we restore the item via the undo stack

zmarinov-astea (approved):

client/src/utils/resource-repo.ts
```diff
@@ -326,16 +326,22 @@ export abstract class ResourceRepo<
     this.update((workingCopy) => {
       toSetInWorkingCopy.forEach((resource) => {
         const resourcePath = [resourceName, resource.id]
+
         // Don't override local changes that may have happened since the push operation started
-        if (get(this.resourceVersions, resourcePath, 0) <= commitVersion) {
+        if (
+          get(this.resourceVersions, resourcePath) !== undefined &&
```
Sachanski: Why do we need this check?

Sachanski (approved):

client/src/utils/resource-repo.ts
zmarinov-astea: Because when it is undefined, it means the resource was deleted locally after sending the request and the response contains the modification of this resource which we should ignore, since we already deleted it. This is what gets the bug solved.

PR #123, opened 2026-09-11
Add IIIF item type, rendered via Clover IIIF (supersedes #77)

dbvisel:
## Summary

- Adds `iiif` as a Tapestry item type. It shows a IIIF image or a full IIIF manifest, with deep-zoom pan and zoom.
- Uses [Clover IIIF](https://samvera-labs.github.io/clover-iiif/) to render manifests, instead of a hand-rolled OpenSeadragon viewer. Clover renders every canvas in a manifest, not just the first one. It shows canvas navigation, metadata, and a table of contents when the manifest has one.
- Supports IIIF Presentation API 2.x and 3.x.
- Accepts two kinds of pasted source: an Internet Archive image-type item URL, or a direct IIIF manifest URL.
- A single-canvas manifest renders through Clover's lower-level `<Image>` component, with tiling forced on. This avoids a real OpenSeadragon bug: a second, simultaneously open instance of the same image style otherwise shows controls but no image. A multi-canvas manifest still uses Clover's full `<Viewer>`.
- This supersedes #77, which added the same item type with a simpler, single-canvas-only viewer. This PR's diff includes #77's changes, since this branch continues directly from it.

## Known limitation

A narrower version of the OpenSeadragon bug above still affects a multi-canvas manifest. If the same manifest is open in two Tapestry items at once, the second instance's first canvas may not render. Every other canvas in it renders correctly, and every other manifest is unaffected.

## Out of scope

- IIIF Collections (a resource that lists many manifests) are not supported yet. Pasting one falls through to Tapestry's existing generic webpage handling. Real picker-based collection import, matching Tapestry's existing bulk-import mechanism, is left for future work.
- A/V (audio or video) IIIF canvases are not supported. This item type accepts only image-bearing canvases, by design.

## Test plan

Manifests and images below are real, live sources, checked working during development.

- [ ] Paste `https://archive.org/details/SPD-SLRSY-1867`. A small Hubble Space Telescope image, from Internet Archive. Confirm it loads and zooms.
- [ ] Paste `https://iiif.wellcomecollection.org/presentation/b18035723`. A 36-page, multi-canvas manifest. Confirm every canvas loads, and canvas navigation works.
- [ ] Paste `https://purl.stanford.edu/fr426cg9537/iiif/manifest`. A Presentation API 2.x manifest. Confirm it loads.
- [ ] Paste `https://iiif.io/api/cookbook/recipe/0024-book-4-toc/manifest.json`. A manifest with a real table of contents. Confirm the structures panel shows real chapter labels.
- [ ] Paste `https://archive.org/details/AMapOfMiddle-earthInTheThird-age`. A large, genuinely tiled image (about 72 megapixels). Confirm deep zoom works.
- [ ] Paste `https://archive.org/details/SPD-SLRSY-1867` twice, into two separate items. Confirm both show the image (the fix for the known OpenSeadragon bug above).


package-lock.json
zmarinov-astea, line 1: Please, checkout the package-lock.json from main, remove the react overrides you placed and run a normal install with clover. We don't want to introduce unwanted behavior or unexpected vulnerabilities by changing the whole package-lock

viewer/vite.config.ts
```diff
@@ -5,6 +5,13 @@ import svgr from 'vite-plugin-svgr'
 
 export default defineConfig({
   plugins: [react(), svgr(), patchCssModules({ generateSourceTypes: true })],
+  resolve: {
+    // @asteasolutions/epub-reader imports plain lodash internally, but declares only
+    // lodash-es. Resolve it to the real dependency instead of installing both.
+    alias: {
+      lodash: 'lodash-es',
```
zmarinov-astea: Remove these aliases from the packages. Rename every lodash import to lodash-es, and remove the lodash installation

core-client/src/components/tapestry/items/iiif/styles.module.css
```diff
@@ -0,0 +1,19 @@
...
+   */
+  --clover-color-primary: var(--theme-text-primary);
+  --clover-color-primary-alt: var(--theme-text-primary-inverse);
+  --clover-color-primary-muted: var(--theme-text-tertiary);
+  --clover-color-accent: var(--theme-background-brand);
+  --clover-color-accent-alt: var(--theme-background-brand-hover);
+  --clover-color-secondary: var(--theme-background-interface);
+  --clover-color-secondary-alt: var(--theme-background-neutral);
+  --clover-color-secondary-muted: var(--theme-background-neutral-hover);
+  --clover-viewer-background: var(--theme-background-primary);
```
zmarinov-astea: The viewer background should be opaque. Also make the image/slide picker shorter (less height)

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+          isTiledImage
+          openSeadragonConfig={OPEN_SEADRAGON_CONFIG}
+        />
+      </div>
+    )
+  }
+
+  return (
+    <div className={styles.root}>
+      <CloverViewer
```
zmarinov-astea: There should be a placeholder here if the item hasn't been activated similar to webpages' click to view placeholder or thumbnail

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+        iiifContent={source}
+        options={{
+          canvasHeight: '100%',
```
zmarinov-astea: Can we display the loading indicator used for webpages here while the clover viewer is loading

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,67 @@
+import CloverViewer from '@samvera/clover-iiif/viewer'
```
zmarinov-astea: should we use a dynamic import for the clover components? You were complaining that the bundle is becoming too large

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,67 @@
...
+  showNavigator: false,
+}
+
+// Do not pass Clover's id or manifestId props. Each one replaces iiifContent instead of
+// identifying the instance.
+export const IiifItemViewer = memo(({ id }: TapestryElementComponentProps) => {
+  const { useStoreData } = useTapestryConfig()
+  const { source } = useStoreData(`items.${id}.dto`) as IiifItemDto
+
+  const { data: manifest } = useAsync(
```
zmarinov-astea: Remove this additional logic for fetching the manifest and displaying a single CloverImage on the canvas. CloverViewer can handle single image iiifs just as well

viewer/vite.config.ts
dbvisel: The bare `lodash` imports are not in our code. They are inside `@asteasolutions/epub-reader`'s own compiled output: `smil/index.js`, `book-loaders/zip.js`, `package-parser/index.js`, and `package-parser/nav-parser.js`. That package declares only `lodash-es` in its own `package.json`, but its published dist files import from `lodash` directly. We cannot rename an import inside a dependency's own published files.

Since `@asteasolutions/epub-reader` is your own package, the real fix belongs there: update those four files to import from `lodash-es`, and publish a new version. Until then, options on our side:

- Keep the alias (`lodash` resolves to `lodash-es` at build time, no extra dependency)
- Use `patch-package` to rewrite the four files at install time
- Add `lodash` back as a real dependency

Which would you prefer for now?

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: This works around a real, reproducible bug, not just a preference. Here is what we found.

`CloverViewer` routes a single-canvas image through OpenSeadragon's `addSimpleImage`, unless the manifest's own `info.json` genuinely restricts `maxArea` below the full image size. Many real IIIF image servers, including Internet Archive's, do not restrict it, even for large images. When that happens, `addSimpleImage` has a real bug: a second, simultaneously open viewer showing that same style of image renders its controls but no image. It stays broken, even after a page reload.

We reproduced this with four different manifests, of different sizes, from two different hosts (Internet Archive and Stanford). Switching the single-canvas path to Clover's lower-level `<Image>` component, with tiling forced on, fixes every one of these. Public OpenSeadragon issues confirm this is a known, unresolved bug class: #2086, #1794, #2700.

We would like to keep this fix. Can we walk through the reproduction together, or is there a different approach you would prefer?

zmarinov-astea: No, it adds too much overhead for not a rare use case. Also the bug will persist for multi-canvas iiifs even with your workaround. So remove the CloverImage logic

viewer/vite.config.ts
zmarinov-astea: Bare lodash is still a dependency. Leave it like that and remove the aliases.

dbvisel: Removed the alias and the workaround. lodash is already a real dependency of @asteasolutions/epub-reader, declared in that package's own package.json. The client build resolves it without help.

core-client/src/components/tapestry/items/iiif/styles.module.css
dbvisel: Fixed. The viewer background is now white and opaque. The thumbnail rail height is now 60px, instead of the default 100px.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Added a placeholder. Before the item is activated, it shows the item's own thumbnail, or an icon if none exists. This matches the webpage item's pattern.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Added the loading indicator. WebpageLoadingSpinner now shows while the lazy-loaded viewer chunk loads. This is the same spinner webpage items use.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Added dynamic imports. CloverViewer now loads through React.lazy. I checked the real build output: its code lands in a separate chunk, not the main bundle.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Removed CloverImage. CloverViewer now renders every manifest, single-canvas or multi-canvas. The known OpenSeadragon bug remains for a second simultaneous instance, as you noted.

package-lock.json
dbvisel: Checked out main's package-lock.json and reinstalled, as asked. This reduced the diff from over 24000 lines to about 2266 lines.

I kept the react overrides. I tested a normal install without them, in a scratch worktree. It still installs two copies of react, 18.3.1 and 19.3.0. Clover IIIF changes how npm hoists react in this tree. Without the override, one copy of react ends up nested under core-client, and useContext then reads null and crashes.

Let me know if you want a different fix for the duplication, or if the overrides are acceptable given the smaller diff.

package.json
```diff
@@ -13,6 +13,10 @@
     "shared",
     "viewer"
   ],
+  "overrides": {
```
zmarinov-astea, line 17: Ok, seems like the issue was caused by "@radix-ui/react-form" dependency which introduced react 18. A better solution is to put "overrides": { "@radix-ui/react-form": "^0.1.16" } instead of directly overriding react. Perform the same checkout from main and install with this override.

core-client/src/components/lib/submit-on-blur-input/index.tsx
```diff
@@ -28,7 +28,7 @@ export function SubmitOnBlurInput({
       onSubmit={(e) => {
         e.preventDefault()
         // We could use a ref as well, but since this is fairly isolated I decided to skip on it --ivo
-        e.target.querySelector('input')?.blur()
+        e.currentTarget.querySelector('input')?.blur()
```
zmarinov-astea: please don't change code that is outside the scope of the current task

core-client/src/components/tapestry/items/iiif/styles.module.css
```diff
@@ -3,6 +3,9 @@
   width: 100%;
   height: 100%;
 
+  /* A deep-zoom image sits on a fixed, opaque backdrop, not a themed one. */
+  background-color: #fff;
```
zmarinov-astea, line 7: background should be set through the options, look at the docs again [https://samvera-labs.github.io/clover-iiif/docs/viewer](https://samvera-labs.github.io/clover-iiif/docs/viewer)

core-client/src/components/tapestry/items/iiif/styles.module.css
```diff
@@ -15,5 +18,8 @@
   --clover-color-secondary: var(--theme-background-interface);
   --clover-color-secondary-alt: var(--theme-background-neutral);
   --clover-color-secondary-muted: var(--theme-background-neutral-hover);
-  --clover-viewer-background: var(--theme-background-primary);
+  --clover-viewer-background: #fff;
+
+  /* Shorten the canvas/slide picker rail (defaults to 100px). */
+  --clover-thumbnail-height: 60px;
```
zmarinov-astea: thumbnails look too stretched now, set only the width to 80px. Can we also modify the slider paddings?

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -1,67 +1,53 @@
...
-          informationPanel: { open: false, renderContentSearch: false },
-          // Some declared search services crash Clover's content-search probe.
-          showMediaSearch: false,
-          withCredentials: false,
-          openSeadragon: OPEN_SEADRAGON_CONFIG,
-        }}
-      />
+      <Suspense fallback={<WebpageLoadingSpinner itemId={id} />}>
+        <CloverViewer
+          iiifContent={dto.source}
```
zmarinov-astea, line 37: Put a loading spinner while the viewer itself loads the iiif image. On both places use the spinner component used when the PDF item gets loaded. We will unify the spinners in a later task

core/src/iiif.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+}
+
+/** Fetch a IIIF manifest. Resolve its first canvas. Return `null` if the manifest is not usable. */
+export async function fetchIIIFFirstCanvas(
+  url: string,
+  signal?: AbortSignal,
+): Promise<IIIFCanvas | null> {
+  const manifest = await fetchIIIFManifest(url, signal)
+  return manifest ? parseIIIFManifest(manifest) : null
```
zmarinov-astea: Can you use manifest parsers from clover's dependencies, not your own parser

client/src/stage/item-factories.ts
```diff
@@ -129,6 +131,41 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
...
+    if ((await getIAItemMetadata(descriptor.item.id))?.mediatype !== 'image') return null
+    manifestUrl = getIAIIIFManifestURL(descriptor.item.id)
+  } else if (mediaType?.includes('json') || /iiif|manifest/i.test(source)) {
+    // A directly pasted IIIF manifest URL. This covers any IIIF source, not just IA.
+    manifestUrl = source
+  } else {
+    return null
+  }
+
+  if (!(await fetchIIIFFirstCanvas(manifestUrl))) return null
```
zmarinov-astea: Here we can just try if the response is parseable with the manifest schema and then proceed

client/src/lib/media.ts
```diff
@@ -62,6 +63,17 @@ export async function getImageItemSize(source: MediaItemSource, width?: number):
...
 
+const DEFAULT_IIIF_WIDTH = 400
+
+export async function getIiifItemSize(source: MediaItemSource): Promise<Size> {
+  const width = DEFAULT_IIIF_WIDTH
+  // An iiif item's source is the manifest URL. Derive the aspect ratio from its first
+  // canvas.
+  const canvas = typeof source === 'string' ? await fetchIIIFFirstCanvas(source) : null
+  const height = canvas?.width ? (width * canvas.height) / canvas.width : width
+  return getClampedItemSize({ width, height })
```
zmarinov-astea: Don't fetch the image to determine the size since it can be arbitrarily large. Use two fixed sizes for the two scenarios - single canvas iiif and multiple canvas iiif determined by parsing the file

server/src/resources/items.ts
```diff
@@ -102,6 +108,25 @@ async function resolveWebSource(item: ItemCreateDto | ItemUpdateDto) {
   item.webpageType = webSourceParser.webpageType
 }
 
+/**
+ * Normalize an iiif item's source. The client usually resolves the manifest itself and
+ * sets `skipSourceResolution`. This function handles the direct-API case instead. It
+ * accepts an Internet Archive item URL or a direct manifest URL. It derives the manifest
+ * URL from an IA URL. It confirms the manifest resolves to an image before accepting it.
+ */
+async function resolveIiifSource(item: (ItemCreateDto | ItemUpdateDto) & { type: 'iiif' }) {
```
zmarinov-astea: I don't think we should resolve the iiif source at all. We should remove this function. It is the client's responsibility to provide the correct data format, as it is with pdfs and e-books

server/src/tasks/thumbnail-generators/index.ts
```diff
@@ -27,6 +28,17 @@ export function hasInherentThumbnail(
...
 ) {
+  if (item.type === 'iiif') {
+    // An iiif item is a deep-zoom tiled image, not a flat one. Render a bounded derivative
+    // of its first canvas from the IIIF Image API instead, and use it as the thumbnail.
+    // The item's source is the manifest URL. Resolve the canvas's image service from it.
+    const thumbWidth = Math.max(MIN_THUMBNAIL_SIZE, item.width)
+    const canvas = await fetchIIIFFirstCanvas(item.source!)
+    if (!canvas) return
+    const derivativeUrl = iiifImageURL(canvas.imageService, { size: `${thumbWidth},` })
+    return generateImageThumbnail(derivativeUrl, { maxDim: thumbWidth })
```
zmarinov-astea: We decided that it would be more intuitive to provide a screenshot of the iiif item with viewer as a thumbnail instead of the iiif source image itself. Proceed with thumbnail generation for iiif items the same way it is done with text and actionButton items for both multi and single canvas iiifs.

package.json
zmarinov-astea: Also put a comment for why this override is here and when should it be removed in the future. We could also open an issue to clover that they need to update their react form dependency.

dbvisel: Done. Overrode @radix-ui/react-form to ^0.1.16 instead of react/react-dom directly. Verified only one react copy installs now (19.2.1). Added a comment in package.json explaining why, and when to remove it. See a6b814a.

dbvisel: Not opened yet. Want me to open it against samvera-labs/clover-iiif, or is this fine to skip for now?

core-client/src/components/lib/submit-on-blur-input/index.tsx
dbvisel: Reverted. It now matches main exactly again.

core-client/src/components/tapestry/items/iiif/styles.module.css
dbvisel: Fixed. Background is now set through Clover's background and canvasBackgroundColor options, not a CSS variable.

core-client/src/components/tapestry/items/iiif/styles.module.css
dbvisel: Fixed. Set only --clover-thumbnail-width now (80px); Clover derives a square from the width alone, so a height caused the stretching. Also tightened the rail's own top padding to match the smaller thumbnail.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Done. Both places (the Suspense fallback and Clover's own loading state, via its customLoadingComponent option) now use the same spinner the PDF item uses.

core/src/iiif.ts
dbvisel: Done. Uses @iiif/parser now (Clover's own dependency for parsing a manifest), via its normalize() function. Added it as an explicit dependency in core, since core can't reach core-client's node_modules where Clover already installs it.

client/src/stage/item-factories.ts
dbvisel: Done. The existence check now just tries to normalize() the response with @iiif/parser and confirms it resolved to a Manifest, instead of hand-parsing it.

client/src/lib/media.ts
dbvisel: Done. Two fixed sizes now: 400x400 for a single canvas, 400x550 for multiple, chosen by parsing the manifest for its canvas count. No image fetch.

server/src/resources/items.ts
dbvisel: Removed. The client resolves the source itself now, the same as pdfs and e-books.

server/src/tasks/thumbnail-generators/index.ts
dbvisel: Done. iiif no longer has a server-side derived thumbnail; it falls through to the same screenshot mechanism as text and actionButton items. Also added a wait for the viewer's manifest and image tiles to load before taking that screenshot, for both single- and multi-canvas manifests.

package.json
zmarinov-astea: Yes, you can open an issue, but no need to reference this project

package-lock.json
zmarinov-astea: Now that we changed the overrides, can you checkout package-lock from main again and run npm install

server/src/tasks/thumbnail-generators/index.ts
```diff
@@ -15,6 +15,7 @@ export interface ThumbnailRenditionOutput {
 
 const MIN_THUMBNAIL_SIZE = 600
 
+// iiif gets a screenshot thumbnail instead (see generate-tapestry-thumbnails.ts).
```
zmarinov-astea: No need for this comment it is self explanatory

server/src/socket/index.ts
```diff
@@ -53,8 +53,8 @@ export class DBSubscriber {
     this.subscriber.notifications.on(DEFAULT_CHANNEL, listener)
   }
 
-  async notify(notification: DBNotification) {
-    return this.subscriber.notify(DEFAULT_CHANNEL, notification)
+  async notify(notification: DBNotification): Promise<void> {
```
zmarinov-astea: no need to specify return type here

core-client/src/stage/renderer/item-renderer.ts
```diff
@@ -112,7 +112,8 @@ export class ItemRenderer<I extends ItemViewModel> extends TapestryElementRender
     super(store, stage, viewModel)
     this.thumbnail = new ThumbnailContainer(null, {
       borderRadius: 8,
-      thumbnailPlacement: viewModel.dto.type === 'image' ? 'stretch' : 'cover',
+      thumbnailPlacement:
+        viewModel.dto.type === 'image' || viewModel.dto.type === 'iiif' ? 'stretch' : 'cover',
```
zmarinov-astea: The new thumbnail generation of iiif items makes it more similar to text items so we shouldn't treat the thumbnail placement as same as the image placement

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+import styles from './styles.module.css'
+
+const CloverViewer = lazy(() => import('@samvera/clover-iiif/viewer'))
+
+// Clover instantiates this itself, with no props, in place of its own plain "Loading" text.
+function IiifLoadingSpinner() {
+  return <LoadingSpinner size="100px" />
+}
+
+// Do not pass Clover's id or manifestId props. Each one replaces iiifContent instead of
```
zmarinov-astea, line 9: this comment is related to the CloverViewer component so move it up there

client/src/pages/tapestry/index.tsx
```diff
@@ -73,7 +72,7 @@ export function TapestryBySlugPage() {
       <Navigate
         to={dashboardPath('home')}
         replace
-        state={{ text: 'Tapestry not found', variant: 'error' } as SnackbarData}
+        state={{ text: 'Tapestry not found', variant: 'error' }}
```
zmarinov-astea: where it said as SnackbarData, and as TapestryIdState replace as with satisfies

client/src/stage/item-factories.ts
```diff
@@ -129,6 +131,41 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
...
+  } else {
+    return null
+  }
+
+  if ((await fetchIIIFCanvasCount(manifestUrl)) === null) return null
+
+  const item = await createMediaItem('iiif', manifestUrl, tapestryId)
+  // The client already resolved the manifest URL, including from an IA source. The
+  // server does not need to redo it.
+  item.skipSourceResolution = true
```
zmarinov-astea: skipSourceResolution is used only for webpage items so remove it from here

client/src/stage/item-factories.ts
```diff
@@ -129,6 +131,41 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
...
+    if ((await getIAItemMetadata(descriptor.item.id))?.mediatype !== 'image') return null
+    manifestUrl = getIAIIIFManifestURL(descriptor.item.id)
+  } else if (mediaType?.includes('json') || /iiif|manifest/i.test(source)) {
+    // A directly pasted IIIF manifest URL. This covers any IIIF source, not just IA.
+    manifestUrl = source
+  } else {
+    return null
+  }
+
+  if ((await fetchIIIFCanvasCount(manifestUrl)) === null) return null
```
zmarinov-astea: For clarity extract a separate function for determining if the manifest parsed correctly. Might use a common helper which returns the parsed manifest itself

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <ItemPlaceholder icon="image" thumbnailSrc={getPrimaryThumbnail(dto.thumbnail)}>
+          Loading…
+        </ItemPlaceholder>
+      </div>
+    )
+  }
+
+  return (
+    <div className={styles.root}>
+      <Suspense fallback={<LoadingSpinner size="100px" />}>
```
zmarinov-astea: WebpageLoadingSpinner looks better for this item. Extract the WebpageLoadingSpnnier in a separate folder and rename it ItemLoadingSpinner. Use it here for the two places where it is needed and also for the webpage items as it is currently used

package-lock.json
dbvisel: Done. Checked out main's package-lock.json again and reinstalled with the react-form override in place.

server/src/tasks/thumbnail-generators/index.ts
dbvisel: Removed.

server/src/socket/index.ts
dbvisel: Removed.

core-client/src/stage/renderer/item-renderer.ts
dbvisel: Fixed. iiif now covers instead of stretches, like text items.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Moved, now right above the CloverViewer declaration.

client/src/pages/tapestry/index.tsx
dbvisel: Done. Both literals now use satisfies instead of as.

client/src/stage/item-factories.ts
dbvisel: Removed.

client/src/stage/item-factories.ts
dbvisel: Done. Extracted isIIIFManifest, built on a new fetchIIIFManifest that returns the parsed manifest. fetchIIIFCanvasCount now also builds on that same helper.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Done. Extracted it into its own component, ItemLoadingSpinner, and use it for the Suspense fallback, Clover's customLoadingComponent, and webpage's existing spinner.

dbvisel:
Fixed a separate bug found while testing: pasting a IIIF viewer's shareable link (not a bare manifest URL) silently imported as a plain webpage instead of a IIIF item, because the old heuristic regex-matched "iiif"/"manifest" against the whole URL, including viewer embed pages.

Now recognizes the standard IIIF Content State API `iiif-content` parameter first, so the manifest URL it points to is used instead. See b9c7ee2.

Sample URL to test with:
`https://projectmirador.org/embed/?iiif-content=https://iiif.archive.org/iiif/b29000427_0001/manifest.json`

client/src/pages/tapestry/index.tsx
zmarinov-astea, line 1: Can you resolve the merge conflicts with astea main?

client/src/stage/item-factories.ts
```diff
@@ -129,6 +131,44 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
...
+/**
+ * Create a IIIF item. Accept three kinds of source: an Internet Archive item URL for an
+ * image-type item, a viewer's own shareable link (e.g. Mirador's embed URL), or a direct
+ * IIIF Presentation manifest URL. Derive the manifest URL from the first two. Confirm
+ * the URL actually resolves to a real manifest before creating the item. A bad or
+ * unrelated URL then falls through to the remaining factories: IA collections/playlists,
+ * then plain webpages. The viewer renders the manifest URL directly and parses the full
+ * manifest itself.
+ */
+const iiifItemFactory: ItemFactory = async (source, mediaType, tapestryId) => {
```
zmarinov-astea, line 140: Remove this huge comment for the iiifFactory, the code is self explanatory

client/src/stage/item-factories.ts
```diff
@@ -129,6 +131,44 @@ export async function createIAMediaItems(tapestryId: string, iaItems: IAItem[])
...
+  if (typeof source !== 'string' || !isHTTPURL(source)) return null
+
+  let manifestUrl: string
+  const descriptor = parseInternetArchiveURL(source)
+  const contentStateUrl = extractIIIFContentStateURL(source)
+  if (descriptor && descriptor.urlType !== 'user-list') {
+    // Only handle image-type IA items here. Let iaFactory handle audio, video, and
+    // collections.
+    if ((await getIAItemMetadata(descriptor.item.id))?.mediatype !== 'image') return null
```
zmarinov-astea: Handle the IA image item import in the IA factory and create the iiif media item for it there

core-client/src/components/tapestry/items/iiif/styles.module.css
```diff
@@ -0,0 +1,29 @@
...
+  --clover-color-accent-alt: var(--theme-background-brand-hover);
+  --clover-color-secondary: var(--theme-background-interface);
+  --clover-color-secondary-alt: var(--theme-background-neutral);
+  --clover-color-secondary-muted: var(--theme-background-neutral-hover);
+
+  /* Shrink the canvas/slide picker rail (defaults to 15rem). Setting a height too
+     stretches the thumbnails, since Clover derives a square from the width alone. */
+  --clover-thumbnail-width: 80px;
+}
+
```
zmarinov-astea, line 25: Remove the .clover-viewer-about-content component padding and add overflow-wrap: anywhere since it overflows.

core-client/src/components/tapestry/items/iiif/styles.module.css
```diff
@@ -0,0 +1,29 @@
...
+.root :global(.clover-viewer-media-wrapper) {
```
zmarinov-astea: Can you reduce the bottom paddings of clover-slider-header and clover-viewer-media-sequence? Also add some small paddings on the sides of clover-viewer-media-sequence. We don't need several .root selectors since selectors can be nested here

core-client/src/components/tapestry/items/iiif/viewer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  // text -- defined here to close over `id`.
+  function IiifLoadingSpinner() {
+    return <ItemLoadingSpinner itemId={id} />
+  }
+
+  if (!hasBeenActive) {
+    return (
+      <div className={styles.root}>
+        <ItemPlaceholder icon="image" thumbnailSrc={getPrimaryThumbnail(dto.thumbnail)}>
+          Loading…
```
zmarinov-astea: Loading is not appropriate text here, maybe "Generating IIIF thumbnail..." or just "Click to view IIIF..."

core/src/data-format/export/v7/index.ts
```diff
@@ -62,6 +62,12 @@ export const ImageItemSchemaV7 = z.object({
   ...V7Props,
   ...actionProps,
 })
+// The iiif type postdates V7. It has no V4-era ancestor to derive from. Its real shape
+// matches an image item, minus the click-action fields.
+export const IiifItemSchemaV7 = z.object({
```
zmarinov-astea, line 65: Remove the comment here for the iiif item schema V7. Also, I think it would be cleaner if you derive it from the book schema.

core/src/iiif.ts
zmarinov-astea, line 1: Remove all comments from this file except the top one and the one on the extractIIIFContentStateURL function

shared/src/data-transfer/socket/types.ts
zmarinov-astea, line 1: Please don't modify this socket file. It is not related at all to the current task

client/src/pages/tapestry/index.tsx
dbvisel: Done. Merged main and resolved the conflict in client/src/pages/tapestry/index.tsx (kept satisfies, kept main's location.state spread).

client/src/stage/item-factories.ts
dbvisel: Removed.

client/src/stage/item-factories.ts
dbvisel: Done. iaFactory now creates the iiif item directly for image-type IA items (falling back to a plain embedded item if the manifest doesn't resolve). iiifItemFactory only handles a viewer's shareable link or a direct manifest URL now.

core-client/src/components/tapestry/items/iiif/styles.module.css
dbvisel: Fixed. Removed the padding, added overflow-wrap: anywhere.

core-client/src/components/tapestry/items/iiif/styles.module.css
dbvisel: Done. Nested all these under .root, reduced both bottom paddings, and added small side padding to the media sequence rail. I used 8px for all of these as a reasonable estimate -- I can't render Clover's info panel interactively from here, so please check these values look right visually.

core-client/src/components/tapestry/items/iiif/viewer.tsx
dbvisel: Changed to "Generating IIIF thumbnail...", matching the webpage item's equivalent placeholder text.

core/src/data-format/export/v7/index.ts
dbvisel: Done. Removed the comment and now derive IiifItemSchemaV7 from BookItemSchemaV7 instead -- book never had the click-action fields image has, so this only needs to omit type, not type/actionType/action.

core/src/iiif.ts
dbvisel: Done. Removed all comments except the top one and extractIIIFContentStateURL's.

shared/src/data-transfer/socket/types.ts
dbvisel: Sorry, reverted -- this was an unrelated formatting change left over from earlier work on this branch, not something from this task.

zmarinov-astea (approved):

PR #124, opened 2026-09-11
Bugfix/tapestry loading

### LINKED ISSUE #118: Tapestry loading animation resets
Shortly after opening a tapestry the animation resets and a brief "Tapestry not found" messages flashes on the screen.  
This happens because the id prop of TapestryLoader switches from '' to the tapestry id, but loading is still false.

### LINKED ISSUE #119: Tapestry loads twice
When navigation a tapestry via the '/t' path, for example /t/b73b1fec-dcbb-43wa-9258-155a3d26c8be, tapestry seems to load, then display for a moment, then loads everything again. This is caused because the state from useLocation() in TapestryBySlugPage gets reset, which causes a reset of the tapestryId and triggers a request to the server to again determine the tapestry Id. We should avoid this request when we have the tapestry Id from the state and also keep the state.

evabanova:
bugs:
[#119 ]
[#118 ]

### DIFF
--- a/client/src/pages/tapestry/index.tsx
+++ b/client/src/pages/tapestry/index.tsx
@@ -7,6 +7,7 @@ import { resource } from '../../services/rest-resources'
 import { dashboardPath, tapestryPath } from '../../utils/paths'
 import { useTapestryPathParams } from '../../hooks/use-tapestry-path'
 import { LoadingLogo } from '../../components/loading-logo'
+import { useRef } from 'react'
 
 interface TapestryIdState {
   tapestryId?: string
@@ -39,7 +40,7 @@ export function TapestryPage() {
       <Navigate
         to={tapestryPath(tapestry.owner!.username, tapestry.slug, mode, location.search)}
         replace
-        state={{ tapestryId: tapestry.id } as TapestryIdState}
+        state={{ ...location.state, tapestryId: tapestry.id } as TapestryIdState}
       />
     )
   }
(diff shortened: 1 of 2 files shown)


client/src/pages/tapestry/index.tsx
```diff
@@ -52,23 +53,31 @@ export function TapestryBySlugPage() {
   const { username, slug, edit } = useTapestryPathParams()
   const mode: InteractionMode = edit === 'edit' ? 'edit' : 'view'
 
+  const cachedTapestryIdRef = useRef<string | undefined>(state?.tapestryId)
```
zmarinov-astea, line 56: I think it would be cleaner to have useState(useLocation().state) instead of the Ref

client/src/pages/tapestry/tapestry-loader.tsx
```diff
@@ -162,14 +162,10 @@ export function TapestryLoader({ id, mode }: TapestryLoaderProps) {
     )
   }
 
-  if (!id || loading) {
+  if (!id || loading || !tapestryDataSync) {
```
zmarinov-astea, line 165: !id and loading all occur when tapestryDataSync is undefined, so we can just leave !tapestryDataSync

client/src/pages/tapestry/index.tsx
```diff
@@ -52,23 +53,31 @@ export function TapestryBySlugPage() {
...
+
+      cachedTapestryIdRef.current = tapestry.id
       return tapestry.id
     },
-    [username, slug, state?.tapestryId],
+    [username, slug],
   )
 
-  if (!loading && !tapestryId) {
+  if (!loading && !(cachedTapestryIdRef.current ?? tapestryId)) {
```
zmarinov-astea, line 80: Here !tapestryId is sufficient since cachedTapestryIdRef.current is being returned by the useAsync

client/src/pages/tapestry/index.tsx
```diff
@@ -63,6 +64,8 @@ export function TapestryBySlugPage() {
         {},
         { signal },
       )
+
+      setState({ tapestryId: tapestry.id })
```
zmarinov-astea: No, don't set the state again since we could return to the same problem. The state is in the dependency array of the useEffect (state?.tapestryId). It is sufficient to just have const [state] = useState((useLocation() ...

zmarinov-astea (approved):

Sachanski (approved):

PR #126, opened 2026-09-14
Bugfix: relationship arrow

evabanova:
Made the line coming out of the arrowhead in a relationship arrow come out of the end of the arrowhead (the base of the triangle), instead of the centre of the triangle, so that the arrow looks better. Added a fixed max of the scale for zoom out.


core-client/src/stage/renderer/rel-renderer.ts
```diff
@@ -108,7 +109,10 @@ export class RelRenderer<R extends RelViewModel> extends TapestryElementRenderer
         isInteractive ||
         (isHoveredElement(pointerInteractionTarget) && pointerInteractionTarget.modelId === id),
       theme: store.get('theme'),
-      relScale: Math.max(1, MIN_REL_VISIBLE_SCALING / discreteScale),
+      relScale: Math.min(
```
zmarinov-astea: This is starting to look too confusing. Can we use this instead? 
const LOCK_VISIBLE_SIZE_AT_SCALE = 0.4
const MAX_REL_SCALE = 5

relScale: clamp(LOCK_VISIBLE_SIZE_AT_SCALE / discreteScale, 1, MAX_REL_SCALE),

core-client/src/stage/renderer/rel-renderer.ts
zmarinov-astea: we don't need to set the Z index here and this constant. We are setting it already in the tapestry renderer.

core-client/src/view-model/rel-geometry.ts
```diff
@@ -63,7 +63,11 @@ export function computeCurvePoints({
   const endpointDistance = distance(from.point, to.point)
 
   function computeSemiCurvePoints({ point, hasArrow }: CurveEndpointParams, direction: Vector) {
-    const curveEndpoint = hasArrow ? translate(point, mul(arrowheadSize / 2, direction)) : point
+    const corner = arrowheadSize / 8
+    const triangleRadius = arrowheadSize / 2
+    const arrowOffset = triangleRadius * 1.5 - corner * (Math.sqrt(2) - 1)
```
zmarinov-astea: Can we extract this math in a function in rel-renderer.ts, since we use it on two places? Something like this `getArrowheadCenterOffset(arrowheadSize: number, arrowheadCornerRadius: number)` 
Also we can extract another function `getArrowheadCornerRadius(arrowheadSize: number) { return arrowheadSize / 8 }`

core-client/src/stage/renderer/rel-renderer.ts
evabanova: I didn't know about this function clamp, pretty neat

core-client/src/stage/renderer/rel-renderer.ts
```diff
@@ -169,13 +169,25 @@ export class RelRenderer<R extends RelViewModel> extends TapestryElementRenderer
...
+    const middle = translate(point, mul(getArrowheadCenterOffset(size, corner), dir))
     const midpoint = new Point(middle.x, middle.y)
 
     graphics
       .roundPoly(midpoint.x, midpoint.y, triangleRadius, 3, corner, angleX(dir) + Math.PI / 6)
       .fill({ color })
   }
 }
+
+export function getTriangleRadius(arrowheadSize: number) {
```
zmarinov-astea: Helper functions should stay at the top of the file, above class/component definitions. Also we can rename this to getArrowheadTriangleRadius to avoid ambiguity

core-client/src/stage/renderer/rel-renderer.ts
```diff
@@ -20,12 +20,13 @@ import { ItemViewModel, RelViewModel, TapestryViewModel } from '../../view-model
...
 import { log } from 'tapestry-core/src/lib/algebra'
+import { clamp } from 'lodash'
 
-const DEFAULT_REL_Z_INDEX = 0
 const LINE_SMOOTHNESS = 0.7
 
 const SCALE_LOG_BASE = 1.3
-const MIN_REL_VISIBLE_SCALING = 0.4
+const LOCK_VISIBLE_SIZE_AT_SCALE = 0.4
+const MAX_REL_SCALE = 15
```
zmarinov-astea: Can we set the max rel scale for 5? This way it will stop growing at zoom level 0.4 / 5 =  0.08, otherwise would be at level 0.026, which is rarely attainable even on large tapestries. You can display the debug panel with the viewport props by passing debug={true} to ViewportDebugData in client/src/pages/tapestry/tapestry.tsx

core-client/src/view-model/rel-geometry.ts
```diff
@@ -63,7 +68,11 @@ export function computeCurvePoints({
   const endpointDistance = distance(from.point, to.point)
 
   function computeSemiCurvePoints({ point, hasArrow }: CurveEndpointParams, direction: Vector) {
-    const curveEndpoint = hasArrow ? translate(point, mul(arrowheadSize / 2, direction)) : point
+    const arrowOffset =
```
zmarinov-astea, line 71: If you agree we can also subtract one pixel here to account for rendering errors or when the rel line gets too twisted.

zmarinov-astea (dismissed):

zmarinov-astea (approved):

Sachanski (approved):

PR #129, opened 2026-09-16
Bugfix: indicator size

### LINKED ISSUE #128: [Bug] Scroll indicators
On very large text boxes - maximum dimensions + 200px font size the scroll indicators become extremely small to the point they are barely visible.

evabanova:
Scaling the font-size (size) of the icon scrolling indicator by the height of the text container. Also scaling the right padding of the text as the indicator grows, so they don't overlap.
Issue: [#128]


core-client/src/components/tapestry/items/text/viewer.tsx
```diff
@@ -129,7 +129,7 @@ export function TextItemViewer({
   }
 
   return (
-    <>
+    <div className={styles.editorWrapper}>
```
zmarinov-astea: No, we don't need this wrapper because it's not a problem if the indicator overlaps with the text. After we click on the text the indicator disappears. Moreover this could mess some styling in other places.

core-client/src/components/tapestry/items/text/styles.module.css
```diff
@@ -17,6 +25,7 @@
     :global(.tiptap) {
       scrollbar-width: none;
       overflow: hidden;
+      padding-right: clamp(16px, 8cqh, 200px);
```
zmarinov-astea: This invisible padding could create some confusion, so I think we are fine without it.

core-client/src/components/tapestry/items/text/styles.module.css
```diff
@@ -3,6 +3,14 @@
   pointer-events: none;
   bottom: 10px;
   right: 5px;
+  font-size: clamp(16px, 8cqh, 200px);
```
zmarinov-astea: Currently the font size of the indicator is 24px, so I think we should stick with that as a minimum. Also I think 3% or 2.5% of the item's longest side is sufficient. See if we will need a maximum in this case.

core-client/src/components/tapestry/items/text/viewer.tsx
```diff
@@ -178,6 +178,6 @@ export function TextItemViewer({
       {!isInteractiveElement && hasScroll && (
         <Icon icon="unfold_more" className={styles.scrollIndicator} />
```
zmarinov-astea: You can pass the scroll indicator size here with a css variable:  style={{ '--scroll-indicator-size': indicatorSize } as React.CSSProperties}, and not use the css cqh unit, but calculate the variable directly from the item dto. You can check out how we pass these variables on other places in the project.

core-client/src/components/tapestry/items/text/styles.module.css
evabanova: Looking at it, I think 2.5-3% is still a bit small, especially for big font size. Maybe leave it 4%? But if you think 3% is ok, I will change it.

core-client/src/components/tapestry/items/text/viewer.tsx
```diff
@@ -176,7 +177,15 @@ export function TextItemViewer({
         {...rteProps}
       />
       {!isInteractiveElement && hasScroll && (
-        <Icon icon="unfold_more" className={styles.scrollIndicator} />
+        <Icon
+          icon="unfold_more"
+          className={styles.scrollIndicator}
+          style={
+            {
+              '--scroll-indicator-size': `${max([24, 0.04 * dto.size.height])}px`,
```
zmarinov-astea: You can use Math.max() instead of lodash's max. Also can you extract these two numbers as constants above? We also want to multiply the ratio to the maximum of two sizes so ratio * Math.max(dto.size.height, dto.size.width), because this could be more indicative of the zoom level when reading.

core-client/src/components/tapestry/items/text/styles.module.css
zmarinov-astea: Ok, but did you try it by taking 3% from the biggest of the two sides? Math.max(width, height)

evabanova: Yes, I tried it now and anyway the change between 3% and 4% is kind of insignificant, so I will just leave it at 3% to be sure.

zmarinov-astea (approved):

Sachanski (approved):

PR #131, opened 2026-09-16
Icon badges on e-book and webpage 

evabanova:


core-client/src/components/tapestry/items/webpage/viewer.tsx
```diff
@@ -116,17 +118,23 @@ export function WebpageItemViewer({
...
+          onLoad={() => setWebpageLoaded(true)}
+          key={`reload-${webpageReloadIndex}`}
+          allowFullScreen
+          allow="autoplay"
+          onPlaybackStateChange={(isPlaying) => dispatch(setItemIsPlaying(id, isPlaying))}
+        />
+      </WebpageLoader>
+      {
+        /* If the webpageType is video content, the video badge should be displayed */
+        !isInteractive && !dto.webpageType && <IconOverlay itemSize={dto.size} icon="globe" />
```
zmarinov-astea: I think it is more appropriate for the icon to be displayed in the WebpagePlaceholder component core-client/src/components/tapestry/items/webpage/placeholder.tsx

core-client/src/components/tapestry/items/book/viewer.tsx
```diff
@@ -216,6 +219,8 @@ export function BookItemViewer({ id, isZipURL }: BookItemViewerProps) {
           )}
         </div>
       )}
+
+      {!isInteractive && <IconOverlay itemSize={dto.size} icon="menu_book" />}
```
zmarinov-astea: Not sure if we want to also display the ItemPlaceholder here and not render the reader if it has not been active similar to the webpages, but we can also skip the icon entirely here. More importantly, you should add these badges to the pixi item renderer `core-client/src/stage/renderer/item-renderer.ts` since in viewer mode the DOM is not displayed at all, only the item's thumbnails are displayed in pixi. You can import an existing tapestry from a ZIP with books and webpages if you don't want to wait for thumbnail generation.

core-client/src/components/tapestry/items/book/viewer.tsx
```diff
@@ -216,6 +222,24 @@ export function BookItemViewer({ id, isZipURL }: BookItemViewerProps) {
           )}
         </div>
       )}
+
+      {!epubState && (
```
evabanova: I am not exactly sure this if this is the right way to do it, but i saw that we use ItemPlaceholder like this is PdfItemViewer.

core-client/src/components/tapestry/items/webpage/viewer.tsx
```diff
@@ -116,17 +116,19 @@ export function WebpageItemViewer({
...
-        src={src}
-        sandbox={sandbox.join(' ')}
-        onLoad={() => setWebpageLoaded(true)}
-        key={`reload-${webpageReloadIndex}`}
-        allowFullScreen
-        allow="autoplay"
-        onPlaybackStateChange={(isPlaying) => dispatch(setItemIsPlaying(id, isPlaying))}
-      />
-    </WebpageLoader>
+    <>
```
zmarinov-astea: We don't need this additional <> </> syntax here

core-client/src/components/tapestry/items/book/viewer.tsx
zmarinov-astea: I think just having 
```
{!hasBeenActive && <IconOverlay... }
```
is ok for now. No ItemPlaceholder and no LoadingSpinner since we need to modify the epub loading logic and the server screenshotting logic also, which will be easier after we merge the IIIF PR

zmarinov-astea (approved):

Sachanski (approved):

PR #132, opened 2026-09-16
Add two possible types of auth 

strendafilov:
Refactor client-side authentication logic to use a Strategy Pattern with a centralized provider registry, eliminating duplicate authentication services.


client/src/config.ts
```diff
@@ -6,8 +6,8 @@ const parsedConfig = deepFreeze(
   z
     .object({
       VITE_API_URL: z.string(),
-      VITE_AUTH_PROVIDER: z.enum(['ia', 'google']).catch('google'),
       VITE_GOOGLE_CLIENT_ID: z.string(),
+      VITE_AUTH_PROVIDERS: z.string().default('google,ia'), // default: both Google and Internet Archive
```
zmarinov-astea, line 10: AI suggests using this schema 
const AuthProvidersEnum = z.enum(["google", "ia"]);

const alternateSchema = z
  .string()                                                      
  .transform((val) => val.split(",").map((item) => item.trim()))
  .pipe(z.array(AuthProvidersEnum))
  .default.....

client/src/components/auth-dialog/styles.module.css
```diff
@@ -0,0 +1,28 @@
+.provider-list {
+  display: flex;
+  flex-direction: column;
+  gap: 14px;
+  padding: 8px 64px;
+  align-items: stretch;
+}
+
+.provider-row {
```
zmarinov-astea, line 9: Can you nest the selectors? The pseudo elements can also be nested
```
.provider-list {
....
   .provider-row {
    ....    
     }
   .divider { 
    .....
     &::before, $::after {
      }
    }
}
```

client/src/auth/google/login-button.tsx
```diff
@@ -1,12 +1,32 @@
...
+  credential: string
+}
 
 export function GoogleLoginButton() {
   const button = useRef<HTMLDivElement>(null)
 
   const type = useResponsive() <= Breakpoint.SM ? 'icon' : 'standard'
 
+  useEffect(() => {
+    window.google.accounts.id.initialize({
```
zmarinov-astea, line 17: This initialize function should be called only once, see the warning [https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.initialize](https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.initialize) That is why we had the auth.prepare functionality. You can instead have some prepare callback in the PROVIDERS_MAP and pass all prepare callbacks to the auth object and store them there, so they get executed together.

client/src/auth/google/login-button.tsx
```diff
@@ -16,6 +36,7 @@ export function GoogleLoginButton() {
         text: 'continue_with',
         size: 'large',
         logo_alignment: 'left',
+        width: 380,
```
zmarinov-astea, line 39: If we remove the width parameter it still looks good I think?

client/src/services/auth.ts
```diff
@@ -112,19 +110,40 @@ export abstract class AuthService<
     }
   }
 
-  async refresh(loadUser = false, signal?: GenericAbortSignal) {
-    await this.doLogin({ authType: 'refreshToken' }, loadUser, signal)
+  async refresh(loadUser = true, signal?: GenericAbortSignal) {
```
zmarinov-astea, line 113: We should better remove the default parameter here and make it boolean. Also for the autorefresh we shouldn't use a lambda function in this case, but a regular one `window.setTimeout((function() { this.refresh(false) }).bind(this), renewAfter)`

client/src/services/auth.ts
```diff
@@ -112,19 +110,40 @@ export abstract class AuthService<
...
+    try {
+      await this.doLogin({ authType: 'refreshToken' }, loadUser, signal)
+    } catch (error) {
+      if (!(error instanceof CanceledError) && loadUser && !this.value.user) {
+        try {
+          await this.doLogin({ authType: 'iaCookies' }, true, signal)
```
zmarinov-astea, line 119: I think this iaCookies login should be handled in the SessionLayout component, which checks if they exist in the browser store. We can also entirely omit this method for now since I don't think its working

client/src/services/auth.ts
```diff
@@ -112,19 +110,40 @@ export abstract class AuthService<
...
+          return
+        } catch {
+          this.update((state) => {
+            state.user = null
+            state.isInitialized = true
+          })
+        }
+      } else if (!(error instanceof CanceledError)) {
+        this.update((state) => {
+          state.user = null
```
zmarinov-astea, line 129: isInitialized is already being set in doLogin. Also why are we setting the user to null here? Are we solving a bug?

client/src/services/auth.ts
zmarinov-astea: If we decide to actually use the iaCookies, then we will need to have them removed when logging out the same way the refresh token is removed.

client/src/auth/google/login-button.tsx
strendafilov: That way, the Google button ends up smaller. I think that when the buttons are aligned it looks better.

zmarinov-astea: Ok, we should also make it look good on mobile

client/src/services/auth.ts
strendafilov: Well, I completely agree that `state.isInitialized = true` is unnecessary. I just thought it was good practice to terminate the user's session when errors occur. Honestly, I can't offer any strong arguments for why setting it to `null` is better, so I'm inclined to remove it.

client/src/auth/google/login-button.tsx
```diff
@@ -2,10 +2,14 @@ import { useEffect, useRef } from 'react'
...
-export function GoogleLoginButton() {
-  const button = useRef<HTMLDivElement>(null)
+interface GoogleLoginButtonProps {
+  isSingleProvider?: boolean
+}
 
-  const type = useResponsive() <= Breakpoint.SM ? 'icon' : 'standard'
+export function GoogleLoginButton({ isSingleProvider = false }: GoogleLoginButtonProps) {
+  const button = useRef<HTMLDivElement>(null)
+  const isMobile = useResponsive() <= Breakpoint.SM
```
zmarinov-astea: rename isMobile to smOrLess since this can get confused with the global exported isMobile variable

client/src/config.ts
```diff
@@ -2,12 +2,24 @@ import { OptionalInt } from 'tapestry-core/src/data-format/schemas/common'
...
+      .filter(Boolean),
+  )
+  .pipe(z.array(AuthProviderEnum))
+
 const parsedConfig = deepFreeze(
   z
     .object({
       VITE_API_URL: z.string(),
-      VITE_AUTH_PROVIDER: z.enum(['ia', 'google']).catch('google'),
       VITE_GOOGLE_CLIENT_ID: z.string(),
```
zmarinov-astea: I think the google client ID can be optional now

client/src/config.ts
```diff
@@ -2,12 +2,24 @@ import { OptionalInt } from 'tapestry-core/src/data-format/schemas/common'
 import { deepFreeze } from 'tapestry-core/src/utils'
 import { treeifyError, z } from 'zod/v4'
 
+const AuthProviderEnum = z.enum(['google', 'ia'])
+const AuthProvidersSchema = z
+  .string()
+  .transform((val) => (val.trim() === '' ? 'google,ia' : val)) // default: both Google and Internet Archive
```
zmarinov-astea: I think z.string().default('google') should be enough here as default handling

client/.env.example
```diff
@@ -1,5 +1,6 @@
 VITE_API_URL=http://localhost:3000/api
-VITE_AUTH_PROVIDER=google
+# Auth providers are specified as a comma-separated list. Supported values: "google", "ia"
+VITE_AUTH_PROVIDERS=google,ia
```
zmarinov-astea: put only google in the example for now

client/src/services/auth.ts
```diff
@@ -53,7 +53,12 @@ export abstract class AuthService<
 
   /** Override in descendants to implement any preparation logic. */
```
zmarinov-astea: We can remove this comment now since we don't have descendants

client/src/services/auth.ts
zmarinov-astea: Yes, we can omit this whole catch block since this logic is being handled in the doLogin function. If the refresh token is incorrect, then a SessionExpiredError is being returned, which should set the user to null

client/src/services/auth.ts
```diff
@@ -83,7 +88,12 @@ export abstract class AuthService<
       const renewAfter = expiresAt - Date.now() - 10_000
       if (renewAfter > 0) {
         clearTimeout(this.autoRefreshTimeout)
-        this.autoRefreshTimeout = window.setTimeout(this.refresh.bind(this), renewAfter)
+        this.autoRefreshTimeout = window.setTimeout(
+          async function (this: AuthService) {
```
zmarinov-astea: Hm, I think we messed this up a little. The correct way to pass the refresh function was to use just `window.setTimeout(this.refresh.bind(this, false)...` You can test this by decreasing the timeout duration.

client/src/services/auth.ts
```diff
@@ -53,7 +53,12 @@ export abstract class AuthService<
 
   /** Override in descendants to implement any preparation logic. */
   protected doPrepare() {
-    // Nothing.
+    if (this.isPrepared) return
```
zmarinov-astea: we don't need the isPrepared prop since this.preparing.state has the same usage and is already being checked by the AuthService.prepare() function below

client/src/auth/index.tsx
```diff
@@ -66,17 +53,19 @@ function RegistrationModal({ initialName }: RegistrationModalProps) {
...
   return (
     <>
-      {createElement(LOGIN_BUTTONS[config.authProvider])}
+      {SingleProviderComponent ? (
+        <SingleProviderComponent isSingleProvider={isSingleProvider} />
+      ) : (
+        <LoginMenu />
+      )}
       {pendingRegistration && (
         <RegistrationModal initialName={pendingRegistration.usernameSuggestion} />
```
zmarinov-astea, line 70: Can you test if the registration flow works correctly with the new changes?

client/src/auth/providers-registry.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface AuthProviderItem {
+  id: string
+  component: React.ComponentType<{
+    onSuccess?: () => void
+    isSingleProvider?: boolean
+  }>
+  prepare?: () => void
+}
+
+const PROVIDER_MAP: Record<string, AuthProviderItem> = {
```
zmarinov-astea: Here for the record type we can use Record<typeof z.infer<AuthProviderEnum>...

client/src/auth/providers-registry.tsx
```diff
@@ -0,0 +1,43 @@
+import { auth } from '.'
+import { config } from '../config'
+import { GoogleLoginButton } from './google/login-button'
+import { IALoginButton } from './internet-archive/login-button'
+
+export interface AuthProviderItem {
+  id: string
+  component: React.ComponentType<{
+    onSuccess?: () => void
```
zmarinov-astea: I'm not sure if we want to have an onSuccess callback, since if the user gets logged in, the LoginButton component unmounts and the dialog gets closed. You can try it out if you want

client/src/config.ts
strendafilov: It is not enough because if we set VITE_AUTH_PROVIDERS= the "" is a valid string. So it has to be   .transform((val) => (val.trim() === '' ? 'google' : val))

client/src/auth/providers-registry.tsx
strendafilov: It also works properly without it

client/src/config.ts
zmarinov-astea: Ok, but if we don't set VITE_AUTH_PROVIDERS at all, we will get undefined. If we want it to always have a value, it would be best to just remove all default handling and release the zod error? :thinking:

client/src/services/auth.ts
```diff
@@ -51,11 +55,7 @@ export class AuthService extends Observable<AuthServiceState> {
     super({ user: null, isInitialized: false, pendingRegistration: undefined })
   }
 
-  /** Override in descendants to implement any preparation logic. */
   protected doPrepare() {
```
zmarinov-astea, line 58: We can also change the protected modifiers to private

zmarinov-astea (approved):

zmarinov-astea (approved):

PR #133, opened 2026-09-23
Add Feature - webpage fullscreen

strendafilov:


client/src/components/tapestry-elements/items/webpage/index.tsx
```diff
@@ -228,10 +267,17 @@ export const WebpageItem = memo(({ id }: TapestryItemProps) => {
   })
 
   return (
-    <>
-      <TapestryItem id={id} halo={toolbar}>
-        <WebpageItemViewer id={id} WebFrame={Webpage} apiRef={apiRef} />
-      </TapestryItem>
+    <div ref={containerRef} className={isFullscreen ? styles.fullscreen : styles.container}>
```
zmarinov-astea: Can we move the display logic to core-client/src/components/tapestry/items/webpage/viewer.tsx ? This is not a client extension and we want it to be usable by the viewer package.

client/src/components/tapestry-elements/items/webpage/styles.module.css
```diff
@@ -15,3 +15,19 @@
     overflow-wrap: anywhere;
   }
 }
+
+.container {
+  width: 100%;
+  height: 100%;
+}
+
+.fullscreen {
```
zmarinov-astea: We don't need the .fullscreen class since the browser applies his own styling when an element is fullscreen

client/src/components/tapestry-elements/items/webpage/index.tsx
```diff
@@ -120,6 +122,23 @@ export const WebpageItem = memo(({ id }: TapestryItemProps) => {
   const [showSaveToWBMPrompt, setShowSaveToWBMPrompt] = useState(false)
   const [isLoadingWBMSnapshots, setIsLoadingWBMSnapshots] = useState(false)
 
+  useEffect(() => {
```
zmarinov-astea: Since we'll use this logic on many places, core-client including, we can extract a hook useItemFullscreen in core-client, which gets the ref and returns both buttons and the isFullscreen flag.

client/src/components/tapestry-elements/items/webpage/index.tsx
```diff
@@ -120,6 +122,23 @@ export const WebpageItem = memo(({ id }: TapestryItemProps) => {
...
+    const onFullscreenChange = () => {
+      setIsFullscreen(document.fullscreenElement === containerRef.current)
+    }
+    document.addEventListener('fullscreenchange', onFullscreenChange)
+    return () => document.removeEventListener('fullscreenchange', onFullscreenChange)
+  }, [])
+
+  const toggleFullscreen = () => {
+    if (!containerRef.current) return
+    if (document.fullscreenElement) {
```
zmarinov-astea: I think we also need to check equality to the containerRef element here since if we are already in fullscreen mode through the bottom tapestry toolbar button, this will get us out

PR #135, opened 2026-09-23
Import tapestry size

evabanova:


client/src/pages/tapestry/view-model/utils.ts
```diff
@@ -173,7 +174,25 @@ export async function insertDataTransfer(
       dispatch(setIAImport(iaImports))
     }
 
-    const viewModels = items.map(createItemViewModel)
+    const customSizedItems = items.map((item) => {
```
zmarinov-astea: The logic for determining the item size should be appended to the existing getWebpageItemSize function in client/src/lib/media.ts  which is used by itemSizes and the createMediaItem function in client/src/model/data/utils.ts

client/src/pages/tapestry/view-model/utils.ts
```diff
@@ -35,6 +35,7 @@ import { Store } from 'tapestry-core-client/src/lib/store'
 import { idMapToArray } from 'tapestry-core/src/utils'
 
 export const DEFAULT_LAYER = 0
+const DEFAULT_LINKED_TAPESTRY_SIZE = { width: 1920, height: 930 }
```
zmarinov-astea: I think EMBEDDED_TAPESTRY_ITEM_SIZE is a better name here?

zmarinov-astea (approved):

